PR #143, opened 2012-04-25
Making impress.js compatible with SVG

csicar:
Why?
Creating the "Presentation-Map" in a vector program is easier than in HTML the data-\* tags can be added there, too.
What was the problem?
SVG-Elements have no Attribute classList, what produces errors. The code adds those Attributes to the SVG-Elements. (I want to add some other things, which makes it even better to use with SVG)

/_Problems with the code-style? -> Tell me a better method of doing it_/

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -163,7 +163,29 @@
         
         return scale;
     };
-    
+    var makeclass = function(elem){
+             if(elem.classList == undefined){
+    		if(elem.className.baseVal != undefined) var s = true;
+    		else var s = false;
+    		elem.classList = {
+    			has: function hasClass(cls) {
+    				if(s) return elem.className.baseVal.match(new RegExp('(\\s|^)'+cls+'(\\s|$)'))
+    				else  elem.className.match(new RegExp('(\\s|^)'+cls+'(\\s|$)'));
+					},
+
+				add: function addClass(cls) {
+   						if (!this.has(elem,cls)) elem.className += " "+cls;
+					},
+
+				remove: function removeClass(cls) {
+   						 if (this.has(elem,cls)) {
+        					var reg = new RegExp('(\\s|^)'+cls+'(\\s|$)');
+        					elem.className=elem.className.replace(reg,' ');
+    					}
+					},
+    		}
+    	}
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
Hi @csicar, in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

csicar:
Well I am still interested in the project and think that the mesioned SVG
support would make use impress.js easier

Fagner Brack <notifications@github.com> schrieb am Mi., 3. Feb. 2016 13:39:

FagnerMartinsBrack:
Thanks for the feedback!

henrikingo:
Hi @csicar 

Could you share a link or two to an example presentation you did in SVG? I'd like to work on integrating your suggestion, but without example content, it's hard to form an opinion on the code, where to put it, etc.

regebro:
Btw, [Hovercraft](https://github.com/regebro/hovercraft) has svg support, and handles that by translating the SVG path to positions and rotations and uses that to generate the impress.js presentation. That might be interesting as an alternative/temporary solution.

henrikingo:
@csicar I'm trying to re-read this thread to understand...

Is the problem here that you would want to add `class="step"` directly to an svg object? Why can't you have a div first, then the svg object inside it?

csicar:
The reason for having `class=step` inside svg is, that it would allow you to create the svg file in an external tool, like inkscape. Inkscape allows you to modify the svg-xml with a build-in editor

henrikingo:
Ah, that sounds interesting indeed. I think I should look into that myself one day!

FWIW, the patch you've proposed here could now easily be implemented as a plugin. It could be simply called `src/plugins/svg`.

henrikingo:
Hi @csicar 

Could you share an example of where you use impress.js over an svg presentation? (presumably together with the above code.) I think your proposal to support svg makes sense, but I'd want to see it working myself in order to commit this code (as a plugin)

henrikingo:
Closing due to lack of response. This still sounds interesting, so you're welcome to open a fresh PR should you want to work on this again.

PR #207, opened 2012-10-27
Added data-substep=""; which will allow for a JavaScript callback to be ...

bendoh:
...used to perform actions upon a slide. Slide will only move forward when the substep function returns true. This allows for animations and the like to be performed on slides before moving on to the next slide

Issue #81, opened 2012-01-26
Add Substeps


gossi:
Hey there,

I would like to see "substeps". I don't know how to call those properly, I give a little description:

Imagine you have your presentation with some slides/steps. Now on one slide you have a list and you want to fly in the list items one-by-one each by a arrow right keystroke. I tried it that way:

```
<ul>
<li class="step">Item 1</li>
<li class="step">Item 2</li>
</ul>
```

It didn't worked, of course I didn't expected that. But would be cool to have something like that.

Thanks very much.

bioinfornatics:
@tehfoo  why you don not do a pull request ?

patricebeutler:
I am working on a solution for this problem. Basically with this solution there is the possibility to add javascript functions as "steps". In this functions you can do whatever you want. Also, there are functions for reverse the changes done by these step-functions. When the work and the testing is done, i will file a pull-request or create a fork.

Announcement:
Tween not steps pls

patricebeutler:
@Announcement Thanks for the hint. My fault. I corrected it.

patricebeutler:
Hey people. If you need this now take a look at my fork: https://github.com/patricebeutler/impress.js. Its exactly the same beside the supsteps-feature.

calebeby:
Any chance this will be merged?

naraesk:
Seems like @bartaz  doesn't have much time for this project recently. However, it would be great if a solution could be found that this repository stays at least maintained. There are several forks of it currently, can't use them simultaneously.

henrikingo:
Hi @calebeby, @naraesk and others

Thanks for pinging on this thread again. Like bartaz, I kinda felt this is a silly powerpoint feature, so it wasn't top of my priority list last year. At the same time, it is true this must be the most requested feature at this point. So I have implemented it [in my fork](https://github.com/henrikingo/impress.js) today. ([Demo](http://henrikingo.github.io/impress.js/examples/classic-slides/#/step-9)).

> it would be great if a solution could be found that this repository stays at least maintained.
> There are several forks of it currently, can't use them simultaneously.

Just to let everyone know, my fork:

* Is actively maintained (for over a year now)
* Intended to be general purpose, so I add feature requests like this, even if I don't personally use them
* I'm happy to take pull requests too (please [familiarize yourself with the plugin framework](https://github.com/henrikingo/impress.js/blob/myfork/src/plugins/README.md))
[...]

PR #214, opened 2012-10-31
innerStep

coltpini:
Edited a different way, I was still having problems with hard returns. So this looks much better.
This defaults to a simple innerStep by just showing the step along the way, but also allows for much more robust stepping for effects similar to, but can go way beyond, keynotes magic move.

basically allowing you to apply css styles on a next() call. There are two ways that this could be implemented and I chose this way, using a json object, because I felt it would be cleaner, though it puts some css in the markup.

The second way this could be implemented is to assign a "step" + i class and have the css be in the css file rather than the tag. Seemed like "6's" to me so I chose the json method because that is what I would personally prefer. 

If you want I could do it the class way. 

If you have questions let me know.

Thanks,
Colt

### DIFF
--- a/index.html
+++ b/index.html
@@ -166,6 +166,10 @@
     -->
     <div id="bored" class="step slide" data-x="-1000" data-y="-1500">
         <q>Aren't you just <b>bored</b> with all those slides-based presentations?</q>
+        <ul style="margin-top: 50px">
+            <li class="innerStep" style="list-style: circle; margin: 30px 0 0 100px;">That go on and on and on.</li>
+            <li class="innerStep" style="list-style: circle; margin: 30px 0 0 100px;">And do basically the same things.</li>
+        </ul>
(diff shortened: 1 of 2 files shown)


FagnerMartinsBrack:
Hi @coltpini, in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

coltpini:
To be honest, I haven't used impress for a long time.

In this pull request was a way, not only to have a sub step, but to also move things around as part of the sub step, kinda like keynotes magic move.

Colt Pini

FagnerMartinsBrack:
@coltpini I am not aware of how keynotes magic move works. Maybe it is worth creating separate issues for each feature?

FagnerMartinsBrack:

In the meantime we will be tracking the substeps functionality on #81

PR #215, opened 2012-11-01
Creation of the multistep (substeps) functionnality

briceparent:
This is the same as https://github.com/bartaz/impress.js/pull/190 , which I closed because the base commit was including irrelevant changes in the indentation.  

Added the substeps / Multistep functionnality that allows to create changes in a single step. Better explained in the example file, between the "positioning-rotating-scaling" step and the "imagination" step. 
It works in both ways (from start to end and from end to start) to enable the users to rewind the presentation.  

As everything in impress.js, it doesn't require javascript skills, only HTML+CSS.  Hope you'll enjoy it!

### DIFF
--- a/css/impress-demo.css
+++ b/css/impress-demo.css
@@ -378,6 +378,39 @@ a:focus {
 }
 
 /*
+This step introduces substeps.
+At first, it only has one line of text, but after 1 action, it adds one line and changes text color to red, and
+after the second action, adds another line another line and changes the color to green.
+This rule hides the added texts
+*/
+#substeps .substep1,
+#substeps .substep2{
+    display: none;
+}
+
+/*
+Those rules set the style for when the sub steps will be shown
+*/
+#substeps{
+    height:400px;
+    -webkit-transition: color 1s;
+    -moz-transition: color 1s;
+    -ms-transition: color 1s;
+    -o-transition: color 1s;
+    transition: color 1s;
+}
+
+/*
+This rule shows the texts and changes the color when the classes are added
+*/
+.substeps_container_1{color: red;}
+.substeps_container_2{color: green;}
+#substeps.substeps_container_1 .substep1,
+#substeps.substeps_container_2 .substep2{
+    display:block;
+}
+
+/*
     The 'imagination' step is again some boring font-sizing.
 */
 
--- a/index.html
+++ b/index.html
@@ -262,6 +262,31 @@
         <p>by <b class="positioning">positioning</b>, <b class="rotating">rotating</b> and <b class="scaling">scaling</b> them on an infinite canvas</p>
     </div>
 
+    <!--
+        
+        This step introduces substeps.
+	
+        Substeps allow a single step to be called more than once in a row, adding a class
+        everytime, which allows to make parts of the page appear or change when the visitor asks to continue
+        the navigation.
+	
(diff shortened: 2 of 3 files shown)


FagnerMartinsBrack:
Hi @briceparent, in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

To give a little bit of context, recently a decision was made in the project to make the development more active and the first task is to clear up older issues like this one to see if the OP is still interested in keep it going.

piranna:
I think this is a basic functionality...

FagnerMartinsBrack:
Hi @piranna, this is indeed a basic functionality and we will try to provide it as soon as possible. The initial plans for impress.js is that it becomes a core mechanism for 3D presentation-like use cases, so this feature might be implemented in a slightly different way in the future.

The reason why I ask if the OP is still tracking this issue is to keep open only one issue related to this feature and close the duplicates :+1:

briceparent:
Hello ! 
I'm not using Impress anymore (not that I've replaced it with any other lib or technology, I just happen not to have the use for it). 
When I opened this, it was just to propose my way of having substeps working like I needed it. I'm not sure if the problem and the solution are still relevant, or if my implementation is the one you need.
Feel free to keep this open and merge it if you feel like so, but don't do it for me as I probably won't find time to help you with it.
Brice

JWorthington:
Here's how I solved it. Rather than load up a whole array of substeps on
load, it looks for children of the activeStep with a class of .substep on
next().
Since there's no hook in the next() function, I replaced the whole
function. I don't know how to formulate this as a pull request, so here it
is.

        // `next` API function goes to next substep within activeStep or to
next step (in document order)
        var next = function () {
// if this step has substeps, and we're not already on the last one, then
"go to" the first/next substep and short-circuit going to the next step
var substeps = $$(".substep", activeStep);
if (substeps.length != 0) {
// if there is an active substep, deactivate it and activate the next
substep
var activeSubstep = $(".substep.active", activeStep);
if (activeSubstep != null) {
// make this one "past"
activeSubstep.classList.remove("active");
activeSubstep.classList.add("past");
// if this is not the last substep, make the next substep "active"
var activeSubstepIndex = substeps.indexOf(activeSubstep);
if (activeSubstepIndex < substeps.length - 1) {
activeSubstep = substeps[activeSubstepIndex + 1];
activeSubstep.classList.remove("past"); // shouldn't be "past", but just to
be thorough
activeSubstep.classList.add("active");
return true; // exit before calling goto(next)
}
// else it was the last substep, so flow through to the normal next-step
behavior
  }
else { // no active substep (yet)
// if not all "past" (i.e., if not any are past), activate the first substep
if ($(".substep.past", activeStep) == null) {
// if none is "active" and none are "past", they must be all future (the
default state), so set the first one active
var activeSubstep = substeps[0];
activeSubstep.classList.add("active");
return true; // exit before calling goto(next)
}
// else all substeps were previously visited, so flow through to the normal
next-step behavior
}
}
            var next = steps.indexOf( activeStep ) + 1;
            next = next < steps.length ? steps[ next ] : steps[ 0 ];

            return goto(next);
        };

FagnerMartinsBrack:
Hi @briceparent thanks for the feedback! In this case we will close this issue in order to track this feature using other similar PRs/Issues. Thank you very much for your effort and for the response.

PR #176, opened 2012-07-04
Added impress:stepgoto event to api

dbankier:
Needed an event to be triggered when going to step (with the target) and prior to the entering it.

PR #234, opened 2012-12-02
Added starttransition event

m42e:
Fire an event on every start of a transition, including the current node (or null if not set) and the next element.

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -292,6 +292,11 @@
                 lastEntered = null;
             }
         };
+		
+        // `onStartTransition` is called whenever a transition to a step element starts
+        var onStartTransition = function (currentStep, target){
+            triggerEvent(target, "impress:starttransition", { current: currentStep, next: target });
+		};
         
         // `initStep` initializes given step element by reading data from its
         // data attributes and setting correct styles.
@@ -485,6 +490,9 @@
             if (activeStep && activeStep !== el) {
                 onStepLeave(activeStep);
             }
+			
+			// trigger that a new transition starts
+            onStartTransition(activeStep, el);
             
             // Now we alter transforms of `root` and `canvas` to trigger transitions.
             //


m42e:
What's the matter with this pull request, I'm about to close it to no longer have it bothering arround in my list ....

FagnerMartinsBrack:
> What's the matter with this pull request, I'm about to close it to no longer have it bothering arround in my list ....

See https://github.com/impress/impress.js/issues/435

FagnerMartinsBrack:
Hi @m42e, in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

m42e:
This is also for my plugin https://github.com/m42e/impress.js-progress

I use it to animate the progress bar (iirc)

FagnerMartinsBrack:
It seems that the fundamental difference between this PR and #176 is that, for #176, the event is triggered the same way as `impress:stepleave`, but with difference that it passes the element, while this PR is supposed to execute the `impress:starttransition` event before each transition starts.

@m42e Can you confirm?

FagnerMartinsBrack:
For future reference, [here is the relevant part of the plugin above](https://github.com/m42e/impress.js-progress/blob/ae023c4c27cef6afed0295ea9908dc4c450a3352/js/impress-progress.js#L17-L19) that uses the `impress:starttransition` event.

henrikingo:
FYI, I've added this in my repo when working on the plugin api and adding the progress bar plugin. It's part of this commit:
https://github.com/henrikingo/impress.js/commit/dd85bbb4437f464d9196830d3b90cfc87b39b49b

I didn't add a new event, rather I simply added `event.detail.next` to the already existing `impress:stepleave` event. That is sufficient for what the progress bar plugin needed.

henrikingo:
`event.detail.next` for `impress:stepleave` events is now also available in master

Issue #232, opened 2012-12-01
Steps as event detail for impress:init


m42e:
Is there a reason not to publish the steps as a detail of the impress:init event detail? I think this will offer many possibilities for extensions and so on. At the moment I'm thinking about somewhat like a progress indicator. Not inside impress.js but as plugin but without having access to the steps order within impress.js I think this is more complicated.

What's your opinion?

Bye
m4n2b

PR #235, opened 2012-12-02
Steps as event detail for impress:init #232

m42e:
see #232

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -399,7 +399,7 @@
             
             initialized = true;
             
-            triggerEvent(root, "impress:init", { api: roots[ "impress-root-" + rootId ] });
+            triggerEvent(root, "impress:init", { api: roots[ "impress-root-" + rootId ], steps: steps });
         };
         
         // `getStep` is a helper function that returns a step element defined by parameter.


m42e:
What's the matter with this pull request, I'm about to close it to no longer have it bothering arround in my list ....

bartaz:
https://youtu.be/M2WTRqCu_DY ;)

FagnerMartinsBrack:
:point_up: Awesome!

Narayanareddy96:
(Y)

bcipolli:
my goodness can someone just press the "merge" button please?

FagnerMartinsBrack:
Your issue does not make it clear why you need this data. Can you share some code on how your plugin is structured in order to assess exactly why this is needed and why you can't pick the data from outside?

Right now I believe you need to access the DOM to be able to get this information "outside" because the markup is the main source of data, that has several performance implications. In the future though, it might be possible to create steps dynamically using an array, in that case you would already have the data without having to expose impress.js internals inside an event handler.

Consider the hypothetical case that you have an external dataset, is it still necessary to expose the steps in the event? Why?

FagnerMartinsBrack:
@m42e Your issue does not make it clear why you need this data. Can you share some code on how your plugin is structured in order to assess exactly why this is needed and why you can't pick the data from outside?

Right now I believe you need to access the DOM to be able to get this information "outside" because the markup is the main source of data, that has several performance implications. In the future though, it might be possible to create steps dynamically using an array, in that case you would already have the data without having to expose impress.js internals inside an event handler.

Consider the hypothetical case that you have an external dataset, is it still necessary to expose the steps in the event? Why?

FagnerMartinsBrack:
Hi @m42e, in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

m42e:
Yes, I'm still tracking this

FagnerMartinsBrack:
@m42e Thank you for the feedback!

Could you please take a look in the [comment above](https://github.com/impress/impress.js/pull/235#issuecomment-155891037)?

m42e:
My intention was to make the steps easily accessible to plugins. E.g. I have written a plugin that shows the current progress (https://github.com/m42e/impress.js-progress).

Why should a plugin do all the selections and so on once again?

FagnerMartinsBrack:
@m42e 

> Why should a plugin do all the selections and so on once again?

Can you elaborate on that? What do you mean by "do all the selections and so on once again"?

FagnerMartinsBrack:
For future reference, [here is the part of the plugin](https://github.com/m42e/impress.js-progress/blob/ae023c4c27cef6afed0295ea9908dc4c450a3352/js/impress-progress.js#L7) that demonstrates the intended usage.

henrikingo:
Just to share my opinion on this. Up in my repo (or is downstream more correct?) I've added the progress bar plugin that is the impetus for this PR. After adding that plugin, as well as another that also needed the list of steps, it is my opinion that this addition is not needed (at the moment).

Since the "root div" is already sent in the `event.target`, and the steps are descendants of the root, it is straightforward and unambiguous to get the list of steps as `event.target.querySelectorAll(".step")`. It does not contribute to making code shorter or clearer to put the steps also separately into the event detail.

(The relevant commit in my repo that I've mentioned: https://github.com/henrikingo/impress.js/commit/dd85bbb4437f464d9196830d3b90cfc87b39b49b )

henrikingo:
Closing based on previous comment.
Note that the `progress` plugin was also merged to master now, thanks @m42e for contributing it!

Issue #238, opened 2012-12-03
Switching Windows With Alt+Tab Sometimes Triggeres Next Step


m42e:
Switching between windows using Alt+Tab triggers next event when switching to browser most of the time. 

I'm not sure how to get rid of this but at least it is mentioned here ;)

olegccc:
I don't know why clicking 'tab' should work for slide change, it is not obvious.

Anyway this workaround can help: 

        document.addEventListener("keyup", function(event) {
            if (event.keyCode === 9) {
                event.stopImmediatePropagation();
            }
        });

(include it before impress.js)

ltregan:
So, here is what is happening.... (at least Windows)

If you are on application A, you press Alt+TAB to switch to the Impress browser. If you are like me, you will release TAB before Alt - at which point the focus goes to the browser with TAB pressed and Alt up.

You then release TAB: the browser receives a keyUp event, and a move to the next slide is triggered. Since TAB is not pressed at that time (remember - it was released before the browser got the focus) checking for event.getModifierState( KeyboardEvent.altKey ) does not help.

The correct solution is thus not to move to next slide upon pressing TAB.

m42e:
maybe a better solution is to memorize if the tab key has also triggered a keydown event. If not, ignore the keyup, else move forward to the next slide.

m42e:
Thanks for the analysis @ltregan .

ltregan:
"maybe a better solution is to memorize if the tab key has also triggered a keydown event. If not, ignore the keyup, else move forward to the next slide." => oh yes, that would work too if we want to keep the TAB key active !

FagnerMartinsBrack:
@m42e @aliva @regebro @diogovincenzi @leroy0211 @Ivanca @eddiemonge @olegccc @ltregan 

Can anybody check if this is still an issue on master?
Also, check [this comment](https://github.com/impress/impress.js/issues/127#issuecomment-4860314). This seems to be a bug with Chrome, can anybody confirm if that is fixed or if there is any open issue for Chrome team?

aliva:
@FagnerMartinsBrack dude, It's been 4 years from my comment, I can't even remember what is impress.js :laughing:

FagnerMartinsBrack:
@aliva Yeah I understand. [See this comment](https://github.com/impress/impress.js/issues/435#issuecomment-177602861) though, we are trying to get this project on it's track again and looking for anyone that is still willing to help.

[...]

PR #247, opened 2012-12-21
a try to solve issue #238

aliva:
in this pull request I tried to solve issue #238

I also removed on direction keys slide change, I think they are a little confusing.


aliva:
meh,

PR #265, opened 2013-03-02
Refactored CSS to encourage non-reuse of 'look-and-feel'; made a template HTML file

truist:
Split the CSS into multiple files, some meant to be reused, some obviously meant not to be.  Also made a template HTML file without all the actual content from the demo, and NOT linking to the "look and feel" bits of the CSS, so that people starting a new project have an easy time getting started, and an extra hurdle if they want to use the "look and feel" CSS.


FagnerMartinsBrack:
Hi @truist, in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

truist:
I have moved on to using reveal.js (https://github.com/hakimel/reveal.js)
but back when I made this pull request, I felt like it would have been a
valuable contribution, and I tried to do it well and carefully - so it
might still be valuable to you. Feel free to use it!

FagnerMartinsBrack:
@truist Thanks for the response! Since you are not tracking this PR anymore and it has conflicts, we are closing it for triage. We will definitely look back to this if the need arises, thanks!

If you think this should be opened later, just ping us here.

PR #266, opened 2013-03-07
add css classes to the body using data-cls

Marvin182:
Hey,

don't know if you want to have this feature in impress.js. I found it very useful in various cases.

Anyway thanks for your awesome work! :)

Marvin

### DIFF
--- a/index.html
+++ b/index.html
@@ -302,7 +302,8 @@
             the element around given axis in degrees; their default value is 0; `data-rotate` and `data-rotate-z`
             are exactly the same;
         * `data-scale` - defines the scale of step element; default value is 1
-        
+        * `data-cls` - defines classes to add to the boly while while viewing the step
+
         These values are used by impress.js in CSS transformation functions, so for more information consult
         CSS transfrom docs: https://developer.mozilla.org/en/CSS/transform
         
--- a/js/impress.js
+++ b/js/impress.js
@@ -309,7 +309,8 @@
                         z: toNumber(data.rotateZ || data.rotate)
                     },
                     scale: toNumber(data.scale, 1),
-                    el: el
+                    el: el,
+                    cls: data.cls ? data.cls.split(" ") : []
                 };
             
             if ( !el.id ) {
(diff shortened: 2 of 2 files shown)


FagnerMartinsBrack:
Hi @Marvin182, in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

Marvin182:
Hi,
I still this this was a missing option, but as I do not longer use impress.js I would not be upset if you decide to discard this idea ;)

FagnerMartinsBrack:
@Marvin182 Thanks for the response!

We can close the PR, but we can keep the idea for later ;D.

That said, do you remember what this feature was about? I don't understand very well the use case for this.

Marvin182:
I wanted to have a way to change the style of the body element in the steps.

FagnerMartinsBrack:
Couldn't you do that with the `stepenter` event? Is this change really necessary?

Marvin182:
I believe not or at least not in a nice, convenient way. But seriously, I told you I do not use impress.js any longer and you already to rejected this 2 year old pull request, why should I waste my time to discuss with you if it would have been necessary or not?

bartaz:
Impress.js currently adds automatic class to body element on each step based on the step id (for example `impress-on-step-5`. This can be used to style body element differently and should be sufficient for most cases.

But of course option to override this automatic class with some custom one can be added too if we find it useful.

FagnerMartinsBrack:
> why should I waste my time to discuss with you if it would have been necessary or not?

Because this discussion can be useful for future developers, just like you thought the code could be useful when you opened the Pull Request 2 years ago.

Anyway, as [per the comment above](#issuecomment-181367766) it seems that now the feature can be accomplished using other existing means.

Thanks for the Pull Request.

Marvin182:
@bartaz The automatic class is sufficient for most use cases, but for some presentations I found it handy to combine and reuse classes and that is what this code can do. I think it is more a design decision, code complexity versus features. Up to you ;)

PR #277, opened 2013-04-08
Enable author to avoid default keyboard bindings for certain elements, like text input elements or code editors.

matix:
Implemented [data-keyboard-capture], so elements inside slides that use keyboard events are not captured by impress.

USAGE:
------

```html
<div id="code-demo" class="step slide" data-x="4400" data-y="2200" data-z="800" data-rotate-x="0">
    <textarea id="code-editor" data-keyboard-capture="true"></textarea>
</div>
```

All keyboard events originating on the textarea, or childs of that element (for example if it gets converted to a CodeMirror instance) are ignored by the impress.js keyboard event handlers.

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -678,14 +678,26 @@
         var api = event.detail.api;
         
         // KEYBOARD NAVIGATION HANDLERS
+
+        function isInteractive (element) {
+            var interactive = Array.prototype.slice.call(document.querySelectorAll("[data-keyboard-capture=true], textarea, input"));
+
+            return interactive.some(function (interactiveElement) {
+                return (interactiveElement == element) || (interactiveElement.contains(element));
+            });
+        }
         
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
Hi @matix, in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

matix:
@FagnerMartinsBrack not really tracking this request, it was something I added to one of my projects and contributed to impress.js.
In retrospective this issue might be solved in a more elegant way than what this PR proposes.

FagnerMartinsBrack:
@matix Thanks for the feedback!

Since you are not tracking this request anymore, then we will close it. We might revisit in the future though, and if you feel that this need to be reopened just ping us out :+1: 

Thanks for the contribution anyway :+1:

PR #280, opened 2013-04-17
simple substep implemenation

misterbrownlee:
Greetings impress.js overlords:

I have made an unobtrusive and clean attempt at providing the basics of a substep feature.  

My implementation follows the syntax and behavior modeled by regular steps:
- denote a substep by adding the css class 'substep' to elements in a step
- init will prepare the substeps with the additional css class 'future'
- next/prev will check within the current step for potential substeps
- navigation through substeps will set 'present', 'active', and 'past' - just like steps
- the only variation is substeps remain 'active' when past
- backward navigation removes 'past/active', restores 'future' as needed
- I've included substep events, following a similar model as regular steps

Because the only presentation-side change is CSS classes, there would be no effect on existing presentations.  A new presentation would have to implement substep specific css to utilize the feature.  This is outlined in the comments of the example html and css. 

I thought it was best to create a separate example, rather than tinker with the existing index.html, as it seems like someone who wants substeps will find the html and figure it out from there.

Substeps are not included as a separate API feature.  I felt this was the right thing because it didn't seem necessary to directly navigate to a substep. I also didn't think deep linking a substep made sense, and the sytax for that would have gotten weird, anyway.  It did make sense to hook into next/prev. If you're adding substeps, you probably wouldn't want prev/next to skip them, even if you're moving via the API.  

I felt an expected default behavior was to have substeps remain active, but that's easily changed.  This could also be easily toggled with a 'data-' property.

The TL;DR notes on my commit are:
- a couple of helper functions for finding and handling substeps
- a small if/else check in next/prev to decide if a substep is available
- a separate example html page with it's own CSS. 

I know substeps are 'powerpointish', and thus subject to controversy.  They also may not meet the impress.js philosophy.  My coding style might not be exactly in-line with the main repo.  If the philosophical issue can be overcome, I'm happy to revise my code as needed to be acceptable.  Just let me know.

Thanks for considering my contribution! 

-- Aaron (aka tehfoo)

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,10 @@ VERSION HISTORY
 
 #### UPGRADING FROM PREVIOUS VERSIONS
 
+In current version calling `impress()` doesn't automatically initialize the presentation. You need to call `init()`
+function from the API. So in a place were you called `impress()` to initialize impress.js simply change this call
+to `impress().init()`.
+
 Version 0.4 changed `goto` API method into `stepTo`. It turned out that `goto` is not a reserved word anymore, so it
 can be used in JavaScript. That's why version 0.5 brings it back and removes `stepTo`.
 
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ It's an (un)fortunate coincidence that a Open/LibreOffice presentation tool is c
 VERSION HISTORY
 -----------------
 
+### 0.5.2 (in development)
 
 ### 0.5.1 ([browse](http://github.com/bartaz/impress.js/tree/0.5.1), [zip](http://github.com/bartaz/impress.js/zipball/0.5.1), [tar](http://github.com/bartaz/impress.js/tarball/0.5.1))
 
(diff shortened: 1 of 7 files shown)


FagnerMartinsBrack:
Hi @misterbrownlee , in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

By the way, this one seems duplicate of #214 and #264.

carljm:
This looks like pretty much the same thing as #264, just with slightly different naming choices. I think it's a good approach in general, and I don't care which one is merged.

FagnerMartinsBrack:

Also, this PR was opened in the dev branch, as from the CONTRIBUTING.md file, it should be opened against the [master branch instead](https://github.com/impress/impress.js/blob/d13428ec226bcb5f6cb918b854507fde1856d042/CONTRIBUTING.md#pull-requests).

PR #281, opened 2013-04-21
"past" and "future" CSS classes can be broken after using goto("#some-id")

frantzmiccoli:
Using goto can break the CSS classes "past" and "future". Here is some code to reproduce the problem on the current demo of impress.js :

    impress().init();
    impress().goto("bored");
    setTimeout(function() {
        impress().goto("ing");
        setTimeout(function() {
            var isPast = document.getElementById("step-2").classList.contains("past");
            console.log(isPast+ " should be true");
    	    var isFuture = document.getElementById("step-2").classList.contains("future");
    	    console.log(isFuture+ " should be false");
        }, 2000);
    }, 2000);

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -401,7 +401,23 @@
             
             triggerEvent(root, "impress:init", { api: roots[ "impress-root-" + rootId ] });
         };
-        
+
+        // `configure` shall be use to define configuration of impress,
+        // configuration is stored in the root node. The only used key right
+        // is "pastfuturepolicy" than can take the value "steps-order" to use the
+        // steps order to affect the "past" and "future" classes on steps.
+        var configure = function (key, value) {
+            root.dataset[key] = value;
+        };
+
+        // `getConfigurationValue` is a helper to access configuration
+        var getConfigurationValue = function (key) {
+            if (!(key in root.dataset)) {
+                return undefined;
+            }
+            return root.dataset[key];
+        };
+
         // `getStep` is a helper function that returns a step element defined by parameter.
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
Hi @frantzmiccoli, in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

From what I see according to the comments in this PR, the change is desired, but then we need first to weight the consequences of breaking existing places that are already using impress.js. Maybe we could instead add an issue to track this problem properly and cross-reference this PR to get detailed information later?

frantzmiccoli:
@FagnerMartinsBrack To my opinion this is still worth pulling, though I haven't been in touch with the most recent code base.

FagnerMartinsBrack:
@frantzmiccoli Thanks for the feeedback. We will revisit this PR later then :+1:

matheod:
It took me some time to understand why my presentation bugged. It was because past and future means viewed/neverviewed.

I read again the demo tutorial, and in fact it was write, but maybe make it even more explicit ?

Also, hopping for something like after/before class :)

FagnerMartinsBrack:
I was wondering why this was closed, then realized that this was closed because the `dev` branch was removed and this PR was created against the `dev` branch.

If [the OP](https://medium.com/@fagnerbrack/what-does-op-means-on-github-8db958b604af) still wants to propose this as [stated on this comment](#issuecomment-181382046), feel free to create the PR against master and open up the discussion again.

frantzmiccoli:
@FagnerMartinsBrack Thanks for this point!

PR #290, opened 2013-05-24
Incorporate / port .skip patch

felixrabe:
Original commit: https://github.com/austinbv/impress.js/commit/f6aaa9f7d9700f362796464a0e3dc88a51bff8eb

Original commit message:

  Skippable positions

  It's nice to be able to remain consistent with the markup for background
  stuff but maybe you don't want it in a slide. Now you can add .skip to an
  element and it will be positioned but not scrolled to. Maybe a little hacky
  but it gets the job done right!

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -427,6 +427,14 @@
                 return false;
             }
             
+            if ( el.classList.contains("skip") ) {
+                var skippedOffset = steps.indexOf(activeStep) > steps.indexOf(el) ? -1 : 1;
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
Hi @felixrabe, in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

FagnerMartinsBrack:

If possible, please provide a gif with a visual example and some usage code with a use case so that we can understand better why this is for exactly. Thanks!

PR #305, opened 2013-06-30
Dev

wojtossfm:
I'm sure this will require quite a bit of discussion before it may or may not be added to the library. The animation plugin (would be) requires quite a bit of work as it currently contains copies of some of the functions from impress.js. This is because I wanted to keep the animation framework separate from the impress.js file. I was hoping to find time to refactor it into something that could be passed into the impress.js init as an option of sorts. This is my first time making a pull request on github so I hope I'm going about this correctly. The substep changes could probably benefit from some codereuse reduction. It may seem slightly chaotic as I got stuck at a certain point and seeing as how I also wanted to finish the presentation. Hope I can clean this up and if you're interested I'd gladly contribute my work to the benefit of the impress.js project.

### DIFF
--- a/js/impress-animations.js
+++ b/js/impress-animations.js
@@ -1,0 +1,484 @@
+(function ( document, window ) {
+	'use strict';
+	
+	// `arraify` takes an array-like object and turns it into real Array
+    // to make all the Array.prototype goodness available.
+    var arrayify = function ( a ) {
+        return [].slice.call( a );
+    };
+	
+	// `$$` return an array of elements for given CSS `selector` in the `context` of
+    // the given element or whole document.
+    var $$ = function ( selector, context ) {
+        context = context || document;
+        return arrayify( context.querySelectorAll(selector) );
+    };
+	
+	// `pfx` is a function that takes a standard CSS property name as a parameter
+    // and returns it's prefixed version valid for current browser it runs in.
+    // The code is heavily inspired by Modernizr http://www.modernizr.com/
+    var pfx = (function () {
+        
+        var style = document.createElement('dummy').style,
+            prefixes = 'Webkit Moz O ms Khtml'.split(' '),
+            memory = {};
+        
+        return function ( prop ) {
+            if ( typeof memory[ prop ] === "undefined" ) {
+                
+                var ucProp  = prop.charAt(0).toUpperCase() + prop.substr(1),
+                    props   = (prop + ' ' + prefixes.join(ucProp + ' ') + ucProp).split(' ');
+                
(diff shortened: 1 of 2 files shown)


FagnerMartinsBrack:
Hi @wojtossfm, in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

wojtossfm:
Hi. I'm not really tacking it actively seeing as how there wasn't much of a response originally. I would need to check how compatible the changes are with the current code base (not sure how much has changed in the meantime). If the proposed changes sound like something you would be interesting in adding to the project then let me know how I can help make integrating the changes into the codebase go smoother.

FagnerMartinsBrack:
@wojtossfm Thanks for the feedback, we will revisit this later thanks.

FagnerMartinsBrack:
As stated in the [contributing guidelines](https://github.com/impress/impress.js/blob/8159876c980d409a85f538ea27d524a5f899328a/CONTRIBUTING.md) it is required that a PR is creating using a "topic branch". Basically it means that a PR should do one thing and one thing only.

In the current state of this PR there are a lot of code changes that will make it harder to review, so I can't say if this s useful or not.

@wojtossfm is it possible that you create a PR splitting each specific change so that we can look each small feature separately? This is not a confirmation that it will be landed, it is just that we need to have a better view of what is being proposed in order to consider this. If you believe it is not possible, then there is no problem, we can close this and you can open a PR later when we have a better plugin structure to be build on top of.

wojtossfm:
Quoting guidelines that were instated long after my PR is a bit... 
surprising.

At the moment I probably won't find time to split it into smaller 
chunks. I can try splitting it in the future although I'm not sure if 
I'll be able to split it into more than 2 chunks - one adding the idea 
of substeps to the original code and a second one being the 
transition/animation extension that utilizes both steps and substeps. At 
the moment then you'll probably want to reject the PR and possibly when 
I'll be able I can try to submit it in separate chunks.

Kinds regards,
Wojciech Michalak

W dniu 2016-03-05 o 05:14, Fagner Brack pisze:

FagnerMartinsBrack:
> Quoting guidelines that were instated long after my PR is a bit... surprising.

Why is that suprising? It is just justifying the need for topic branches, which was not documented clearly earlier.

> At the moment then you'll probably want to reject the PR and possibly when I'll be able I can try to submit it in separate chunks.

@wojtossfm Thanks for the response! We will be closing this for now then but feel free to submit in the future when we hope to have a better structure to build on top of :+1:

PR #307, opened 2013-07-02
replace tap left/right for touch devices with swipe left/right

tshedor:
Pros: allows full-width HTML overlays to presentation, like a navigation bar, on touch devices
Cons: threshold may be variable, not as reliable as tap on left or right of screen; requires two event listeners

Wanted to provide this as an alternative to the core tap left/right, although in all honesty, tap is the better solution in most situations.

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -761,22 +761,27 @@
             }
         }, false);
         
-        // touch handler to detect taps on the left and right side of the screen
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
Hi @tshedor, in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

FagnerMartinsBrack:

PR #351, opened 2013-11-29
Upgrade to typographer quotes

dvdrtrgn:
Looks even classier with "curly" quotes, doncha think?

### DIFF
--- a/index.html
+++ b/index.html
@@ -164,7 +164,7 @@
         
     -->
     <div id="bored" class="step slide" data-x="-1000" data-y="-1500">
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
This seems reasonable. Ping @impress/mergers.

henrikingo:
Heh. +1 Since no code is changed.

FagnerMartinsBrack:
I would argue that for the sake of consistency it should be changed for all apostrophes, but then it will be hard to maintain and will make no difference since this PR refers to the ones that are shown in the demo.

Pierstoval:
No problem for me as it's in the demo, so it's ok

PR #360, opened 2014-01-06
Impress.js script now runs when page load is complete

eduardrock:
Added a window.onload event for all the impress.js script, 
so now you can load the impress.js script tag from head. There's no need to load it from the end of the body tag.

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -21,6 +21,10 @@
 
 // You are one of those who like to know how things work inside?
 // Let me show you the cogs that make impress.js run...
+
+// Al this script loads only when the complete page is loaded, so you can load the Impress.js <script> tag
+// from the <head> without problems and avoid loading it from the end of the <body> tag
+window.onload = function(){
 (function ( document, window ) {
     'use strict';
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
@ericnelsonaz The semicolon in the end is not superfulus, if it is included via `script` tag and another `script` tag right below uses an Immediately Invoked Function Expression (`(function () {}());`), then it will interpreted as an argument for the previous Function Declaration if the semicolon is omited:

```javascript
window.onload = function() {
  // This is immediately executed and pass the IIFE result as an argument.
}(function () {}());
```

FagnerMartinsBrack:
@eduardrock The only thing impress.js does is to provide the API, how it will be loaded is the responsibility of the developer. It doesn't make sense coupling impress.js into a single way to load the script by default because a user might want to load it in a different way.

FagnerMartinsBrack:
Closing due to the lack of feedback from the OP. If you want to work in the PR again, feel free to comment here to open up the discussion again

PR #402, opened 2014-06-10
exposing, impress.roots, an window.Impress constructor and Impress.reset helper

akhoury:
The `roots` object is not exposed, hence it's not very useful if one wants to delete an existing instance. I understand that it's not very practical to have multiple instances of an "impress" object on a single page, it's almost impossible to handle, since each instance manipulates the `location.hash` value and listen to its changes, and I personally agree, and I think you should just have a single static instance, unless you want to support a light router, i.e. `#/:rootId/:step-index`

However, since you're already keeping track of each instance, why not construct them the *conventional* way, (no way is the right-way in JS) using a `new Impress( rootId )` syntax, and why not exposing a helper for one to reset or delete one or all instances. 

My use case consists changing the impress DOM template, then reloading the deck, without refreshing the page.

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -190,12 +190,7 @@
     }
     
     // GLOBALS AND DEFAULTS
-    
-    // This is were the root elements of all impress.js instances will be kept.
-    // Yes, this means you can have more than one instance on a page, but I'm not
-    // sure if it makes any sense in practice ;)
-    var roots = {};
-    
+
     // some default config values.
     var defaults = {
         width: 1024,
@@ -217,7 +212,7 @@
     // It's the core `impress` function that returns the impress.js API
     // for a presentation based on the element with given id ('impress'
     // by default).
-    var impress = window.impress = function ( rootId ) {
+    var Impress = window.Impress = function ( rootId ) {
         
         // If impress.js is not supported by the browser return a dummy API
         // it may not be a perfect solution but we return early and avoid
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
Hi @akhoury , in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

By the way, this change introduce breaking changes for those who are already using impress.js. We need to weight the cost-benefits of that.

akhoury:
I don't have a strong opinion on this - feel free to proceed as you see fit. Although it would be nice to have a way to refresh the deck without the need to refresh the page.
Thanks for asking.

FagnerMartinsBrack:
I have created an issue to track the use case: #517. We might revisit this later if someone bring this up again. Thanks for the feedback!

js/impress.js
```diff
@@ -639,10 +637,42 @@
...
+
+        body.className.split(' ').forEach(function(className) {
+            if (className.indexOf('impress-on-') === 0)
+                body.classList.remove(className);
+        });
+        location.hash = '';
+    };
+
+    // alias to support the backward compatible impress().init()
+    window.impress = function( rootId ) {
```
akhoury: @FagnerMartinsBrack btw, what does this PR break, as far backward compatible? I did set a global `impress` function

FagnerMartinsBrack: That was definitely my mistake. I didn't realized that the code was still exposing the `impress` global here =(

akhoury: no worries

akhoury:
that works. Thanks

PR #412, opened 2014-07-27
onStepLeave should fire before current slide loses "active" class

andygoblins:
Conceptually, I expect the step that calls onStepLeave to be "active" when onStepLeave runs. This allows me to use onStepLeave as a teardown function when leaving a slide.

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -426,6 +426,11 @@
                 // presentation not initialized or given element is not a step
                 return false;
             }
+
+            // trigger leave of currently active element (if it's not the same step again)
+            if (activeStep && activeStep !== el) {
+                onStepLeave(activeStep);
+            }
             
             // Sometimes it's possible to trigger focus on first link with some keyboard action.
             // Browser in such a case tries to scroll the page to make this element visible
@@ -480,11 +485,6 @@
             }
             
             var targetScale = target.scale * windowScale;
-            
-            // trigger leave of currently active element (if it's not the same step again)
-            if (activeStep && activeStep !== el) {
-                onStepLeave(activeStep);
-            }
             
             // Now we alter transforms of `root` and `canvas` to trigger transitions.
             //


FagnerMartinsBrack:
Hi @andygoblins , in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

andygoblins:
I'm still tracking, and I'm still interested

FagnerMartinsBrack:
@andygoblins Thanks for the feedback! We will put this in the queue for review.

FagnerMartinsBrack:
@andygoblins Can you please provide a real world example where this feature might be useful?

andygoblins:
One example would be "exit" animations. A script acting as a onStepLeave hook can access the current slide as .active to animate/modify the slide's elements before the slide transition.

I have also used it to have "sticky" headers that float above the rest of the content:

```
<h1 id="stickyheader" style:"position:fixed; top:10px;visibility:hidden">This header floats above the content on slides 1 and 2</h1>
...
document.addEventListener("impress:stepenter", function(event) {
    if document.querySelector("#slide1.active") != null {
        document.getElementById("stickyheader").style.visibility="visible";
    }
});
/* This part only works if my patch is in place: */
document.addEventListener("impress:stepleave", function(event) {
    if (document.querySelector("#slide2.active") != null) {
        document.getElementById('stickyheader').style.visibility="hidden";
    }
});
```

henrikingo:
Hi @andygoblins 

I think the idea behind this patch is a bit misguided, and therefore I think this patch is both unnecessary and the changing of semantics is a bad idea. Feel free to defend your case more strongly, but let me explain what I see when I read this:

First of all, events are asynchronous. So even if you would trigger the event when the element being left is still `.active`, there's no guarantee it will remain `.active` until your event handler executes. (I suppose implementations of javascript are such that it actually will always be the case, but it is still bad programming style to assume that it will be.)

The `impress:stepleave` event already has the div being left in `event.target`. You don't need to find it with `querySelector()` or otherwise.

impress.js sets a class `impress-on-<id>` on the body element when it enters a new step. Hence, what you're doing above with a floating header, you can do purely in CSS without any JavaScript:

    h1#stickyheader {
        visibility: hidden;
    }
    body.impress-on-slide1 h1#stickyheader,
    body.impress-on-slide2 h1#stickyheader {
        visibility: visible;
    } 

Some other teardown activities you might want to do in javascript, but for reasons explained above, you in any case shouldn't rely on it being executed before or after `impress:stepleave`, nor do you need the `.active` class to find the element you want.

The most important reason is that impress.js at this point is much too mature to change such semantics. There are probably at least hundreds of users that rely on the semantics being exactly as they currently are, and changing it would cause more harm to them than it would benefit you.

FagnerMartinsBrack:
> The most important reason is that impress.js at this point is much too mature to change such semantics. There are probably at least hundreds of users that rely on the semantics being exactly as they currently are, and changing it would cause more harm to them than it would benefit you.

Definitely! This is definitely the main point why it is so hard to change anything in impress.js without very good coverage. I would say thousands rely on this and even changing any part of the CSS API in a major version will cause a lot of breakage, and since people can't easily test their presentation, we would need a very good changelog listing all breaking possibilities. It is so much work that the effort in increasing the coverage and API docs would be more worth it than listing all breaking possibilities.

If we can fix this use case using the API as is, we could mark this as closed so that we can focus in the cases that can't be fixed the way we have the API right now.

@andygoblins Can you please take a look in @henrikingo's point and see if we can mark this as solved with the current API that we have? Thanks.

andygoblins:
I agree.

PR #425, opened 2014-09-01
Add Relative Positioning

SethGunnells:
So I went back and looked through the pull requests and found that this was suggested and implemented a while back. My implementation of it is a little different though so I'm curious to see what you all think.

It works by allowing users to add a attribute called `relative` to the root element(s) to enable relative positioning. Then, if they should need to use absolute positioning on an element, they can do so by adding the `absolute` attribute.

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -246,7 +246,7 @@
         
         // current state (position, rotation and scale) of the presentation
         var currentState = null;
-        
+
         // array of step elements
         var steps = null;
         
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
Hi @SethGunnells , in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

SethGunnells:
I haven't been tracking this, no. As I stated in the first comment, I became aware that this feature had, essentially, been already implemented and was pending a merge after I submitted my version. I haven't been keeping up with the project since. You can drop this if you like. Or pull it in. If you want to keep it, I'll gladly help in whatever way is needed.

FagnerMartinsBrack:
As from [this comment](https://github.com/impress/impress.js/pull/485#issuecomment-198634252) we will be closing the PRs related to this feature due to the low amount of interest and maybe take a look in the code later when we jump into it. For now we will be keeping track of it on #478.

Thank you very much for the PR @SethGunnells!

PR #442, opened 2014-12-01
Update index.html

matthiasn:


FagnerMartinsBrack:
Hi @matthiasn it seems that the word is correct, it means that despite HTML and CSS you need to have design skills too.

matthiasn:
Sure, but then it would be "you need to be a designer, too".

FagnerMartinsBrack:
Wang to update the PR? :)

FagnerMartinsBrack:
Closing due to the lack of feedback from the OP. If you want to work in the PR again, feel free to comment here to open up the discussion again

PR #452, opened 2015-01-16
"too" becomes "to"

maliktunga:
It's a mistake, "you need to be a designer" and not "you need too be a designer".


maliktunga:
This pull request is a duplicate of https://github.com/bartaz/impress.js/pull/432

Issue #435, opened 2014-10-22
State of the project


FagnerMartinsBrack:
Hi @bartaz, is this project going to be actively developed?

In several threads you stated no time for working into this project and the latest commits were just regarding grammars.

Impress is impressive (!!), and there are many useful pull requests waiting to be landed. Any news if this project is going somewhere or if you are willing to handle it to someone else to continue active development?

Why do I ask this? Well, maybe I am not the only one, but its sad to see fragmented projects with useful features that could be centralized. If there is any chance that this project is going to get on track, it would be nice to disclose the information, so one may not jump into creating another repository and potentially reinvent the wheel.

Also, I noticed there is a lack of general documentation, so it is hard to know exactly what is supported and what is considered an implementation detail (although the code is pretty well documented and one can have a very good basis).

Anyway, with > 1k watchers and > 22k stars it is unlikely not to exist at least one willing to take over or be an active collaborator. Is there any news regarding the project state? Just to get into the context.

Thanks.

FagnerMartinsBrack:
My current project is not just about a regular slide presentation, it should be scalable and impress.js does not seems to be the best fit for it.

I switched to [jmpress](github.com/jmpress/jmpress.js), a jquery plugin that handles advanced features that impress currently doesnt, with the potential to develop plugins for it. Since I am using angular, I also created an [angular-jmpress prototype](https://github.com/web-stories/angular-jmpress). It is just a prototype of  a basic usage, since I didn't find any repository containing an angular port of jmpress.js or impress.js.

Any help is appreciated. jmpress need contributions and angular-jmpress is just something that, if evolved, could be eventually ported to the jmpress organization.

Pierstoval:
When I'll have more sliders to make, I'll try jmpress thanks for the notice @FagnerMartinsBrack  :)

regebro:
You probably meant to link to https://github.com/jmpressjs/jmpress.js

It's demo page is completely broken on Firefox and Linux, unfortunately, I might look at it when they fix that. :-)

FagnerMartinsBrack:
@regebro Yeah, I fixed the link.

By the way, Firefox is pretty broken for impress itself, at least on Windows. A few days ago Firefox 34.0 wasn't working, several artifacts in the screen when changing a step. It is related to a known issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1098266.

[...]

PR #457, opened 2015-02-21
Feature: Substeps

patricebeutler:
We needed the "substep"-feature mentioned in the
https://github.com/bartaz/impress.js/issues/81 issue.

We implemented 4 API functions for assigning functions to the steps.
Every mentioned API function takes two parameters: The id of the step
and a function that should be executed on a given time.

There can be one function per step for entering.
There can be one function per step for leaving.
There can be many functions per step that act as "substeps". The user
can then modify everything he wants with simple js. These functions are
called trough next(). The first defined function is executed first.
There can be one function per step that allows resetting the changes
done by the "substep-functions".

Our changes are all backwards-compatible, so it will not break anything
from anyone. We have tested it with a variety of browsers.

We did this with 13 line html (demo), 10 line css (demo) and 46 lines js
(framework). In total less than 70 lines of code added.

If you have something you want to speak about or you are not happy with,
feel free to discuss.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
+This is a fork of impress.js. It supports substeps, but besides that, its the same as the original. The link of the original is: https://github.com/bartaz/impress.js.
+
 impress.js
 ============
-
 It's a presentation framework based on the power of CSS3 transforms and 
 transitions in modern browsers and inspired by the idea behind prezi.com.
 
@@ -12,7 +13,7 @@ impress.js may not help you if you have nothing interesting to say ;)
 HOW TO USE IT
 ---------------
 
-[Use the source](http://github.com/bartaz/impress.js/blob/master/index.html), Luke ;)
+[Use the source](https://github.com/patricebeutler/impress.js/blob/master/index.html), Luke ;)
 
 If you have no idea what I mean by that, or you just clicked that link above and got 
 very confused by all these strange characters that got displayed on your screen,
@@ -30,7 +31,7 @@ EXAMPLES AND OTHER LEARNING RESOURCES
 
 ### Official demo
 
-[impress.js demo](http://bartaz.github.com/impress.js) by [@bartaz](http://twitter.com/bartaz)
+Look at the html-file in this repo. It is slightli different than the original from bartaz.
 
 ### Examples and demos
 
@@ -46,17 +47,6 @@ on the wiki, too.
 There is also a book available about [Building impressive presentations with impress.js](http://www.packtpub.com/building-impressive-presentations-with-impressjs/book) by Rakhitha Nimesh Ratnayake.
 
 
-WANT TO CONTRIBUTE?
----------------------
-
-If you've found a bug or have a great idea for new feature let me know by [adding your suggestion]
-(http://github.com/bartaz/impress.js/issues/new) to [issues list](https://github.com/bartaz/impress.js/issues).
-
-If you have fixed a bug or implemented a feature that you'd like to share, send your pull request against [dev branch]
-(http://github.com/bartaz/impress.js/tree/dev). But remember that I only accept code that fits my vision of impress.js
-and my coding standards - so make sure you are open for discussion :)
-
-
 
 ABOUT THE NAME
 ----------------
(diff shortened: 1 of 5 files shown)


bartaz:
Hi @patricebeutler, thanks for the commit. I'll have a closer look later.

But can you shortly explain to me how different it is from current `step:enter` and `step:leave` events?
They are also a functions that you can bind to steps. I'm pretty sure currently it is not possible to implement substeps using these events, but maybe with some small changes in how these events work it would be possible?

I'm thinking that maybe if there is a possibility to 'cancel' leaving the step inside of the `step:leave` handler and show a substep instead it would work?

What do you think? Would it be possible to turn your solution into using built in events?

patricebeutler:
Thanks for your quick response.

At first we implemented the two functions for directly handling substeps (addActionToStep and addResetActionToStep). This functions are not events at all and only are used for the handling of the substeps.

At this point everything worked fine. As you noticed, there are also the two functions "addEnterActionToStep" and "addLeftActionToStep". These functions were implemented for usability-concerns and integrity. It is correct that these are like a wrapper for the impress:stepenter and the impress:step:leave events (with the difference they take the id of the step for binding them directly to a step).

With these function you can, as a user, write (pseudocode)

    addEnterActionToStep("someStepId", function(){
        // Some code here
    });

instead of

    addEventListener("impress:stepenter", function(){
        if (step.id == "someStepId"){
            // Some code here
        }
    });

Or in different words: We thought it would be cool for the user to have this possibility. But still, if the user wants to execute code when each and every step is entered, the "impress:stepenter"-event still would be needed.

If you think the "addEnterActionToStep" and the "addLeftActionToStep" are not required for the user, it is easy for us to remove them from the api. The substeps-feature would still work, but the user would have to use the "old" way to bind "stepenter" and "stepleave". This is not awful, but it will result in more lines of coding for the user, since these events are used more as soon as you have substeps. Also, we dont know if every user of impress is able to subscribe to an event and then use an if-statement to determine if the event was fired on a specific slide or not (you know your userbase better than us).

Let me know what you think, thanks again. We are happy to do changes as you like and then file a pull-request again.

bartaz:
You could attach the event to a given step directly to avoid checking the id inside:

```
document.getElementById('someStepId').addEventListener("impress:stepenter", function(){
    // ...
});
````

This may mean that `addEnterActionToStep` and `addLeftActionToStep` are quite obsolete as they do the same thing as events.

But of course it doesn't solve the substeps part.

patricebeutler:
Oh I see now how this works. Our fault. Perfect, the removal of these functions will make the code even smaller. We will remove it in a few hours and then file a new pull-request if that works for you. Then you can do the review and decide if its matching your visions of impress.

bartaz:
Sure. I think if you the fix in same branch it will automatically appear in this pull request, so new one shouldn't be needed.

patricebeutler:
Very good. This is my first time doing this in github, so I didn't know. Thanks for the information. Let me know what you think about the code. Thank you very much.

FagnerMartinsBrack:
Hi @patricebeutler , in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

patricebeutler:
First of all, we are very excited to hear that there is new fire in this project!

I can't answer with a simply yes or no: Yes, we are still tracking this. We strongly think that substeps are one of the most useful features and we still heavily use the pull-requested version today. Still, we aren't sure this is the best way of doing it: At the time of implementing it, our time was very limited, which resulted, from our view, in the smallest possible result: Substeps can be programmed in javascript (this gave maximum flexibility while keeping things simple). There is just one drawback: Everything besides the substeps can be used declaratively.

We submitted the pull-request because the project was very inactive at that time and we thought that this may help some folks. But since there is activity in this project again, I think you should first discuss the best way of achieving it (if you decide to implement substeps). If you conclude that substeps should be created programmatically (and not declaratively), this pull-request does just that (and it's already in use by a few people). So maybe you will be able to use the (or parts of the) code, If you want declarative substeps or something more fancy, I suggest you close the request.

Thank you for your work!

FagnerMartinsBrack:
@patricebeutler Thanks for the feedback!

This PR seems to have been opened against the `dev` branch and it contains a lot of unrelated commits. We will probably revisit this if we need ideas on how to implement substeps.

Thanks!

PR #459, opened 2015-02-25
compatable with old-browser

dongseok0:
Some old browser, in my case android JB stock browser, is passed compatible check.
But element.dataset gives empty string if there is no matching value, and it makes toNumber function's fallback not work so all default styles values are set 0.
With this fix, it works well if I remove mobile checking.

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -82,7 +82,7 @@
     // it into a number. If it is not possible it returns 0 (or other value
     // given as `fallback`).
     var toNumber = function (numeric, fallback) {
-        return isNaN(numeric || undefined) ? (fallback || 0) : Number(numeric);
+        return isNaN(numeric) ? (fallback || 0) : Number(numeric);
     };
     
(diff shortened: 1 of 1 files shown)


bartaz:
Thanks for reporting.

But I think that with your fix `toNumber(0)` will not work, because it will treat `0` as `undefined`.

dongseok0:
I thought `toNumber()`'s argument should be string type? `toNumber('0')` works fine.
If it's not, maybe need to test with [this](http://stackoverflow.com/a/1830844/1853949).

bartaz:
If accessed directly from data attribute it would be a string, but at least before your change it was also working fine for numbers.

But maybe using `isNaN(parseFloat(numeric))` as suggested by article you mentioned would help.

dongseok0:
Yeah, but `isNaN(parseFloat('4f'))' will not work. I think it needs to use more precise `isNumeric()` function. I'm gonna make another PR!

PR #469, opened 2015-03-16
Updated bower.json and removed duplicate main entry

matb:
bower.json did contain "main" entry twice. This leads to incorrect behaviour.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,12 +1,11 @@
 {
   "name": "impress-js",
   "version": "0.5.3",
-  "main": "./js/impress.js",
   "description": "It's a presentation framework based on the power of CSS3 transforms and transitions in modern browsers and inspired by the idea behind prezi.com",
   "homepage": "https://github.com/bartaz/impress.js",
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
Hi @matb , in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

matb:
Hi @FagnerMartinsBrack, personally I don't have a preference. But this should definitively be fixed because it causes issues in combination with other tools.

FagnerMartinsBrack:
@matb Could you please give an example of which tools this may cause problem in jsfiddle or jsbin?

matb:
Sorry for the late reply. I just checked the current version of bower.json and the issue has been already resolved there. This one can be closed.

FagnerMartinsBrack:
Actually it is not, see https://github.com/impress/impress.js/blob/master/bower.json, there are 2 bower `main` entries.

FagnerMartinsBrack:
I have created an issue to track this, see #522

PR #475, opened 2015-04-25
Adds an improved mouse wheel event listener

michaelneu:
Including the `impress-scoll.js` allows you to use a wireless mouse with or without a flawless mouse wheel sensor instead of a presenter.

### DIFF
--- a/index.html
+++ b/index.html
@@ -360,6 +360,7 @@ if ("ontouchstart" in document.documentElement) {
     Of course you can wrap it in any kind of "DOM ready" event, but I was too lazy to do so ;)
     
 -->
+
 <script src="js/impress.js"></script>
 <script>impress().init();</script>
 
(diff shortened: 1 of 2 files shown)


FagnerMartinsBrack:
Hi @minedev , in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

minedev:
Hi, to be honest I haven't tracked this request in a while as there was basically no progress on the other PRs as well. 

I just checked the current demo of impress.js and noticed that this feature is not implemented yet, but could actually come in handy, so I'll track it again. Maybe I'll rewrite it soon, as I now think some parts are quite hacky.

FagnerMartinsBrack:
@minedev Thanks for the feedback! If you want to make sure the problem does not happen then I suggest not relying in the demo, but checking out master and test locally. Although, I still think this feature was not implemented.

Anyway, since you are still tracking we will come back to review this later then, Thanks!

FagnerMartinsBrack:
This is a possible duplicate of https://github.com/impress/impress.js/pull/51. @michaelneu can you please check if the problems discussed there still applies for this PR? Thanks!

FagnerMartinsBrack:
By the way, this PR was opened against `dev`. The new [contributing guidelines](https://github.com/impress/impress.js/blob/8159876c980d409a85f538ea27d524a5f899328a/CONTRIBUTING.md) states that it should be opened against `master`, can you fix that too?

FagnerMartinsBrack:

Issue #478, opened 2015-05-25
Feature request: relative data-* values


bendavis78:
I think it'd be nice to be able to position slides relative to the previous one. 

So, for example, instead of this:

``` html
<div class="step slide" data-x="0">Slide A</div>
<div class="step slide" data-x="1000">Slide B</div>
<div class="step slide" data-x="2000">Slide C</div>
<div class="step slide" data-x="3000">Slide D</div>
```

We would have this:

``` html
<div class="step slide" data-x="0">Slide A</div>
<div class="step slide" data-x="+1000">Slide B</div>
<div class="step slide" data-x="+1000">Slide C</div>
<div class="step slide" data-x="+1000">Slide D</div>
```

That way, if I needed to insert a slide somewhere in there, it doesn't mess up the overall flow.

henrikingo:
Hi @bendavis78 

There are a couple pull requests that implement this in various ways. I have one here: https://github.com/henrikingo/impress.js/

jacekkopecky:
I too have various extensions including relative positioning at https://github.com/jacekkopecky/impress.js

henrikingo:
@jacekkopecky Wow that's an impressive number of commits!

We need to get all this stuff into one repo. I'll try to make progress with @bartaz tomorrow or next week, see if we can poke this forward a bit.

jacekkopecky:
And it's not all - I still have some things I've used but haven't pushed yet.

I actually just thought about starting over, refactoring impress to better suit my needs; but I'll still push everything I've done and used to my fork and I'll be very happy if any of it is useful in the main repository. Let me know and I can prepare PRs.

henrikingo:
Oops, I've completely forgot to come back to this thread.

@jacekkopecky Some months ago already, I've added [a Relative Positioning Plugin](https://github.com/henrikingo/impress.js/blob/myfork/src/plugins/rel/README.md) to my for of impress.js, where I implemented a plugin API based on bartaz' guidance. Please have a look and let me know what you think.

[This demo presentation](http://henrikingo.github.io/impress.js/examples/classic-slides/) uses the relative positioning data- attributes.

henrikingo:
The relative positioning plugin is now part of impress.js master, as of 1.0.0-beta1

PR #473, opened 2015-04-20
relative position

Sneezry:
Set relative position based on pre step. Add `data-relative=“true”` to
set relative position. See demo with index-with-relative-position.html.
All translate, rotate and scale are relative.

PR #131, opened 2012-03-30
allow positioning relative to previous step

nchauvat:
This makes it easier to start at origin of canvas and move at each step relative to the previous step. Use data-x="+=1000" instead of absolute value, for example.

PR #485, opened 2015-07-03
Relative positioning of steps

henrikingo:
Congratulations for creating gh://impress/impress.js! To kickstart the activity, I'm submitting this pull request.

I'm aware that there's discussion of creating a plugin architecture for impress.js. Although simple, these changes could be implemented as a "positioning plugin", which would convert the relative positioning values to absolute positions (as in the core code). In fact, I'm also willing to work on creating such a positioning plugin api, but will need guidance on what type of api/plugin semantics we need to adhere to.

***

Adds support for following new attributes for steps:

    data-rel-x
    data-rel-y
    data-rel-z

These can be used to specify position instead of the absolute position arguments `data-x/y/z`.

The value of `data-rel-x/y/z` is added to `data-x/y/z`. If not specified, also the `rel-*` values are inherited from the previous step. This makes it easy to create a boring presentation where each slide shifts for example 1000px down from the previous.

In addition, adds another demo under examples/classic-slides/. In addition to demoing relative positioning, this demo is also intended to be a much simpler (perhaps boring) and beginner friendly introduction than the existing flashy demo by @bartaz.


FagnerMartinsBrack:
Does this have any relationship with https://github.com/impress/impress.js/pull/473, https://github.com/impress/impress.js/pull/425 or https://github.com/impress/impress.js/pull/131?

henrikingo:
Yes. They're all alternative implementations of the same basic need.

There seem to be 2 general approaches:
 1. Add an attribute to the root element saying something like `positioning="relative"` and then use the same `data-x`, `data-y` and `data-z` attributes, just that the numbers have different meaning. (This should be extended with a possibility to define relative/absolute separately for each step.)
2. Add new attributes, like I have done above: `data-rel-x` and so on. #131 is also in this general category, though different.

I suppose both of these have their own merits. I like second approach for being clearer, not overloading same attribute with meanings that could change. For example if you copy some HTML from a presentation with relative positioning, to another with absolute positioning, you'll have a mess with (1), but it would just work with (2).

All of these call for a "Positioning API", which would allow implementing these as well as more elaborate positioning (like automatically distribute steps along an SVG path...). If an API exist, both types of plugins can coexist, although I think this is quite a core feature (even if a plugin!) so we should also show some leadership and suggest one markup style as the "official" choice. (We can change our minds later if users seem to prefer some other plugin.)

henrikingo:
Oh, I should note that the addition of examples/ and in particular impressConsole.js extends the scope of this pull request quite a lot and its quite improper of me to put unrelated changes into same branch. Basically they reflect the fact that I saw this project as unresponsive and therefore have added things to the branch that were useful for me. The pull request for relative steps must in any case be rewritten to use an api, so from that point of view this should be ignored. 

(But I suggest these are not closed, rather we should respond to all that this kind of functionality will be added via API, and people are welcome to propose such an api, and to do so there must be test coverage first.)

bartaz:
Yes, there are couple of issues/PRs about relative positioning.

My preference was different - I didn't want to introduce any new attributes and use existing `data-x` etc with a different notation (similar to what jQuery has for animations). So for relative positioning you would write `data-x="+=100` or `data-rotation="-=45"`. Having `+=` or `-=` in the beginning of the value would mean that position should be calculated based on previous step.

But I guess discussion which solution is better and which could go into impress.js is something we can have later. For sure this is one of the most requested features.

FagnerMartinsBrack:
> But I suggest these are not closed, rather we should respond to all that this kind of functionality will be added via API, and people are welcome to propose such an api, and to do so there must be test coverage first.

I would argue that the related PRs could be closed for triage if there are too many duplicates or with lots of unrelated changes, there is no value in keep open PRs that cannot be landed as is. One important thing though, is not to forget about important features like this one by maintaining an open issue specific to the purpose of solving this kind of problem **with cross-reference to the proposed PRs to consult later**.

For an example take a look in the issue https://github.com/impress/impress.js/issues/516 and the linked PRs.

henrikingo:
@FagnerMartinsBrack: You're right. Closing current PRs and opening single issue for discussion makes more sense.

@bartaz: Yes, that approach definitively has some merits. It avoids any ambiguity (e.g. you could safely copy paste html into another context) but also avoids adding new attributes. Note that in general I would expect some new positioning schemes to definitively add new attributes, such as positioning along an SVG curve.

FagnerMartinsBrack:
I have closed all other PRs due to the lack of feedback from the OP, except for #425. That one seems to be more easy to review and the OP has demonstrated interest in keep it going. Should we close this in favor of #425 instead? Or, if the amount of changes are going to be reduced for this PR, should we keep this one and close the other?

Despite which one we keep, it will only be landed once we have a proper structure for testing anyway. I just want to focus in only one PR instead of keeping duplicates open.

henrikingo:
Also, either of these has to be rewritten as a plugin, not touching core code.

FagnerMartinsBrack:
@henrikingo So what do you suggest, close this and the other one in favor of #478? It might be worth reducing the amount of open stuff so that we can focus in what can be potentially done in the current state of the project. I guess that landing stuff in the core now is not a problem, since we will probably have to change the whole structure in upcoming releases anyway.

FagnerMartinsBrack:
> So what do you suggest, close this and the other one in favor of #478?

@henrikingo I will assume we should close both and follow the issue #478 instead, as you mentioned above:

> @FagnerMartinsBrack: You're right. Closing current PRs and opening single issue for discussion makes more sense.

PR #487, opened 2015-07-06
add 2 API function

yalishizhude:
demo

showPage：
An API function used to display pages info like this: '13/26' at the bottom by adding footer element.
```impress().showPage() ``` to make it work.
showProgress：
An API function used to display progress bar at the top by adding header element.
```impress().showProgress() ``` to make it work.


yalishizhude:
These two APIs maybe can make the presentation more easily understandable and professional ^_^

FagnerMartinsBrack:
It's not the responsibility of impress.js to format the data. I would say that impress could keep track of the steps position, but I am not sure if this is useful enough to be exposed as an API. You can easily keep track of your steps using an external model. For example, if you are writing your html by hand, you know the position beforehand, so you can print without impress.

henrikingo:
I have to say I disagree with the previous comment. If you have 99 slides, and numbered them manually, and then add a slide to the beginning, you would have to renumber all of them. Which would be silly, when you can do what this pull request proposes instead.

Currently the main problem is that no pull requests are being handled by anyone. There's been discussion of opening up this project, and moving into a more modular and then more inclusive type of project. If that actually happens soon, this pull request would definitively fit in well.

FagnerMartinsBrack:
> If you have 99 slides, and numbered them manually, and then add a slide to the beginning, you would have to renumber all of them

It makes sense to conveniently expose the position as a `Number` or as a `position` Object Literal. But not to format the data, that's the client job.

henrikingo:
Ah, I agree with that much. I wasn't looking too closely at this specific implementation. Such a discussion makes more sense once we have an architecture and process to submit pull requests against. (I have my own pull requests waiting, and fully expect to have to rewrite it against some refactored impress.js before it gets accepted.)

bartaz:
hi @henrikingo @FagnerMartinsBrack, thanks for this discussion.

You are both right. As @FagnerMartinsBrack said - I personally don't see such paging/progress functionality in the core of impress.js. Especially that it can be already achieved with a little bit of JS trickery and counting slides.

And for sure I'd like to see impress.js more flexible and open for pull requests. Just at this point (without more flexible architecture) any additional feature will turn code into bigger spaghetti.

Unfortunately I don't think I'll be able to make it on my own as I don't have time for that and even idea how impress.js should be refactored to make it all possible. I have maybe a very blurry view of how I wish this code could be organised, but it's not enough to make it real.

Maybe we could just create an issue here for refactoring, have some discussion there and if anyone would be able to implement it just make it real? What do you think?

henrikingo:
Hi @bartaz 

Yes, please create such an issue. Even if you don't have time to do it, I think the initiative for discussion needs to start from you. Or, if you don't even want to initiate discussion, you need to simply pick a person/persons who you delegate to. Based on #435 all the rest of us have been waiting for a cue from you and nobody will propose anything before there is a signal that you're open to proposals. (In fact, there was some GSoC project that we were waiting for?)

As for myself, I would currently have time to contribute into refactoring, but need someone else, who is more familiar with current state of the art in JavaScript, to dictate the coding style, style of plugin api, "dependency injection", unit testing frameworks and so on... Once I'm told which tools to use, I can contribute my ideas on how to refactor impress.js and also do refactoring, create new api's, etc.

henrik

FagnerMartinsBrack:
I have made a comment in that issue with some suggestions for the next steps to make: https://github.com/impress/impress.js/issues/435#issuecomment-132184917

FagnerMartinsBrack:
@yalishizhude Thanks for the PR!

Closing due to the comments above and also due to the lack of feedback from the OP. It seems that this feature is not useful for the current state of the project and it can be easily done using external data.

PR #496, opened 2015-09-25
Added better swiping support for touch devices.

and3rson:
So, recently I've been doing some stuff for a presentation and noticed that it is not working on my Android.

A little examination has shown me that Android, iPhone and iPad are blacklisted for the swipe.

I think the time has come to disable the blacklist as most modern (and not-so-modern) devices nowadays work flawlessly with Impress.JS.

So, I've implemented a "swipe" left & right effect and removed "Android" User-Agent from the blacklist.

Check out the video below.

[Impress.JS swipe demo](https://www.youtube.com/watch?v=9MgFKaHCONY&feature=youtu.be)

Hope this is useful to somebody :)

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -84,7 +84,21 @@
     var toNumber = function (numeric, fallback) {
         return isNaN(numeric) ? (fallback || 0) : Number(numeric);
     };
-    
+
+    var toNumberAdvanced = function (numeric) {
+        if (!(typeof numeric == 'string')) {
+            return toNumber(numeric);
+        }
+        var ratio = numeric.match(/^([\d\.]+)([wh])$/);
+        if (ratio == null) {
+            return toNumber(numeric);
+        } else {
+            var value = parseFloat(ratio[1]);
+            var multiplier = ratio[2] == 'w' ? window.innerWidth : window.innerHeight;
+            return value * multiplier;
+        }
+    };
+
     // `byId` returns element with given `id` - you probably have guessed that ;)
     var byId = function ( id ) {
         return document.getElementById(id);
@@ -299,9 +313,9 @@
             var data = el.dataset,
                 step = {
                     translate: {
-                        x: toNumber(data.x),
-                        y: toNumber(data.y),
-                        z: toNumber(data.z)
+                        x: toNumberAdvanced(data.x),
+                        y: toNumberAdvanced(data.y),
+                        z: toNumberAdvanced(data.z)
                     },
                     rotate: {
                         x: toNumber(data.rotateX),
@@ -555,7 +569,11 @@
             
             return goto(prev);
         };
-        
+
+        var interpolate = function(a, b, k) {
+            return a + (b - a) * k;
+        };
+
         // `next` API function goes to next step (in document order)
         var next = function () {
             var next = steps.indexOf( activeStep ) + 1;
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
It would be amazing if there was an algorithm to simulate exactly the changes a native do when swiping (like moving the step while you drag until a given threshold).

and3rson:
@FagnerMartinsBrack I think that's possible to do having transition interpolations for mid-frames. The only question is performance. I will try to experiment with that :)

and3rson:
@FagnerMartinsBrack I've added interpolation for frames to provide native-like dragging. The performance seems usable on mobile devices :) Check this demo out:

[Impress.JS swipe demo](https://www.youtube.com/watch?v=LXkwQ5s7VSY&feature=youtu.be "Impress.JS swipe demo 2")

P. S. I also changed `toNumber` to `toAdvancedNumber` in some slides to allow using values like:

```html
<div class="slide step" data-x="0.5w" data-y="2h"> ... </div>
```

This means to use **0.5 * screen width** for X and **2 * screen height** for Y.

FagnerMartinsBrack:
Really awesome!

I believe though, for the sake of usability, that it should swipe using the pointer as the reference instead of the actual slide edges.

and3rson:
@FagnerMartinsBrack Thanks :) However I'm not sure if I understand you correctly, can you describe the desired behavior?

FagnerMartinsBrack:
>  can you describe the desired behavior?

In the current implementation, the step change only occurs if the border of both slides pass through the middle of the screen. If you start swiping from the end, the pointer is after the middle of the screen, so the swipe doesn't happens.

Example below:

impress_swipe_example

and3rson:
@FagnerMartinsBrack ah, I see now. Sure, I can just change threshold to some lower value (currently 1/2 screen width, probably need 1/4 or 1/5 or smth)

and3rson:
@FagnerMartinsBrack Done! Check out the demo site :)

FagnerMartinsBrack:
:+1:

and3rson:
Anyone with write access to the repo around? :)

FagnerMartinsBrack:
I am not sure if this is going to be merged soon, see #435.

The reason I suggested some improvements was because I see a lot of value in something like this, even if it is not merged. I will probably cherry-pick some of the changes if you don't mind ;D

and3rson:
@FagnerMartinsBrack use as you wish, it's open for everyone :)

FagnerMartinsBrack:
@and3rson I find this feature very useful. Just pinging to say that we haven't forgotten it. The problem is that without proper automated testing this can't be landed. We will keep this in the queue and come back later.

To give a little bit of context, recently [a decision was made in the project](https://github.com/impress/impress.js/issues/435#issuecomment-177602861) to make the development more active. If you want to keep track of what we are doing, please check #512.

and3rson:
@FagnerMartinsBrack Awesome! Thanks for info.

You guys are doing an awesome job.
In case you get back to this feature later, I'll be glad to fix or reimplement it to suit the new architecture (and create tests as well.)

henrikingo:
Hi @and3rson 

Your code is really cool! I've merged this in the past 2 commits to [my fork|http://henrikingo.github.io/impress.js/]. You can try it with the [official demo here|http://henrikingo.github.io/impress.js/]. Note: To play with the swiping, you may want to stop the autoplay by clicking on the pause button to the bottom right.

Testing with my 3 year old tablet, the official demo animates a swipe pretty smoothly on Chrome. Only the "tiny ideas" slide is tough, as it zooms (ie scales) so heavily. On Firefox for Android it was quite sluggish, but doing a quick swipe instead of trying to "enjoy it" will bring you to the next step no problem.

Details:
 - I made the w/h positioning as a separate commit. It's also a bit different: Since there was already a relative positioning plugin in my fork, I naturally added this to the same plugin. As a result, a presentation needs to use `data-rel-x="2w"`, etc...
 - I separated the swiping code into a new [impress().swipe()|https://github.com/henrikingo/impress.js/blob/c44fd0f4c102e98a90ee564b91a5cf4f66cec6c3/src/impress.js#L652] api method, and a [touch plugin|https://github.com/henrikingo/impress.js/blob/c44fd0f4c102e98a90ee564b91a5cf4f66cec6c3/src/plugins/touch/touch.js]. That worked out pretty well, the `swipe()` method just needs a single parameter to render the swipe: a ratio between -1.0 to +1.0.

Pierstoval:
Is there any news from here? 😉

FagnerMartinsBrack:
@Pierstoval This depends on https://github.com/impress/impress.js/issues/575. We can't start implementing new features until we have a considerable amount of integration test coverage in the project. We don't want to break the presentations of who is using it :)

henrikingo:
This is now merged to master (with some refactoring to fit the new plugin based approach). Thanks @and3rson, I enjoyed your code! (Never did anything that swiped before.)

and3rson:
@henrikingo Glad it helps!

PR #497, opened 2015-09-28
Fixing small typo

g-akshay:


bartaz:
Thanks, but I actually meant 'notation' – 'notation' of rotation value is `data-rotation="90"`. :)

PR #509, opened 2016-01-14
Shortcuts for overview

mn22nw:
I found it really useful  to have shortcuts for the overview when creating my presentation. 

If the user doesn't have a slide/div with the id overview, nothing get's messed up. It just doesn't show the overview.   

I know it's not the absolute best solution out there. But this is not too bad either I think.
By having a div with the id overview, you can go in and change the values for how big the overview is and where it's supposed to be placed.  When I added a lot of slides I found that the existing scale,  x and y values weren't optimal for my presentation.  Hence, it's good to let the user define the overview, and this makes it easy for them to do so.


js/impress.js
```diff
@@ -716,12 +716,19 @@
                     case 40: // down
                              api.next();
                              break;
+                    case 36: // character o
+                    case 79: // home 
+                            var overview = document.getElementById("overview");
+                            if (overview) {
+                                api.goto(overview);
+                            }
+                            break;
```
jacekkopecky: the above could just be `api.goto("overview")` which will do nothing if such an element is not found.

mn22nw: Thanks! I've changed it now. :)

bartaz:
This pull request has too many unrelated commits in it.

Issue #494, opened 2015-09-12
Pluggable architecture for impress.js


henrikingo:
In some other issues, @bartaz has explained that he would like impress.js grow in functionality by the current core remaining much the same, but allowing for plugins that provide extended functionality. Lots of people have lots of ideas how they'd like to add to impress.js. In this writeup I've tried to collect all ideas (aka requirements) I've heard of, and then propose a high level architecture to meet those. In the end I've listed things that need to be discussed, aka I need help with those.

User stories
------------

I've heard people wanting to do following things with impress.js

 * I want to do presentations that are nicer than PowerPoint slides
   * I like writing presentations in HTML
   * I want to create a GUI tool for creating impress.js presentations
   * I need to be able to use this PowerPoint replacement without internet 
     connectivity
   * I want it to be something I can install as an application on my laptop
 * I want to write my presentations in some ohter markup language (MarkDown, 
   RST) and compile that into impress.js
 * I want to embed 3d animations in my web page or app
 * I want to embed impress.js functionality into my product (e.g. as a Drupal 
   module)
 * I want to extend the HTML notation used to create presentations (e.g. 
 relative positioning)
 * I want to add some standard visual elements to my presentations (e.g. 
 back & forward buttons)
 * I want impress.js to ship with a repository of nice templates I can base my 
 presentations on
 * I want to share my presentation to remote viewers over HTTP, while I control 
 which step they see
 
Note that implementations of most of the above already exists as separate 
projects or pull requests, which is how we know about them.
 
Architecture diagram
--------------------
 
We want to keep the core impress.js pretty much doing what it does today, but 
also create plugin APIs to allow for extensions that implement higher level
features without bloating the core. In fact, some of the above requirements
would be contradictory: a user wanting a big ass PowerPoint replacement wants
completely different functionality from someone wanting to embed a 3D animation
in his CMS software.

Taking the requirements above, we get an architecture like in the attached 
SVG diagram. Ahem, so github doesn't support SVG, here's a PNG, I'll try to share the original somewhere later.

![impress-architecture-diagram](https://cloud.githubusercontent.com/assets/2482018/9831089/bf3a713a-5951-11e5-9fa5-b13f995885f0.png)

For discussion
--------------

Following decisions need to be made by a person more familiar with current
state of art in JavaScript than myself:

 * How are plugins and APIs created (coding style, how are plugins loaded, 
   dependency injection...)
 * Repositories: Is every plugin a separate github repository? If yes, how 
[...]

PR #510, opened 2016-02-01
Adding qunit test framework and first set of basic tests.

henrikingo:
This is the beginning of having a test framework in impress.js.

Part of issue #494


FagnerMartinsBrack:
You need to add your Github e-mail address to the `user.email` git config in order for Github to find the correct author of the commit.

test/index.html
```diff
@@ -0,0 +1,250 @@
...
+  <script src="helpers.js"></script>
+  <script>
+    // In case tests didn't complete, we are left with a hash/anchor pointing somewhere. But we want to start from scratch:
+    window.location.hash = "";    
+    // I don't understand exactly why, but we must call init() before the actual tests
+    impress().init();
+    
+    QUnit.test( "Initialize Impress.js", function( assert ) {
+      assert.ok( impress, 
+                 "impress declared in global scope" );
```
FagnerMartinsBrack: In lieu of a formal styleguide, it would be better to be consistent with how the current codebase is written. In this case there is no need for a line break between arguments.

FagnerMartinsBrack: The same goes for the other assertions

test/index.html
```diff
@@ -0,0 +1,250 @@
...
+// This is the one that doesn't work. Test 17 fails if this is done inside a test()
+//      assert.deepEqual( impress().init(), undefined,
+//                        "impress().init() called." );
```
FagnerMartinsBrack: Shouldn't these kind of comments be written as PR comment instead of being inside the code?

test/index.html
```diff
@@ -0,0 +1,250 @@
...
+    // Note, it's clearly implied in the impress.js code, that when we support
+    // plugins, the keypress api should be a plugin. Then this test of course
+    // should go there as well.
+    QUnit.test( "Keypress API", function( assert ) {
+      var done = assert.async();
+
+    });
+
+    // Cleanup
+    QUnit.test( "Cleanup", function( assert ) {
```
FagnerMartinsBrack: Could this be done with the `before`/`after` APIs?

test/index.html
```diff
@@ -0,0 +1,250 @@
...
+    
+
+
+
+
+
+    // Note, it's clearly implied in the impress.js code, that when we support
+    // plugins, the keypress api should be a plugin. Then this test of course
+    // should go there as well.
+    QUnit.test( "Keypress API", function( assert ) {
```
FagnerMartinsBrack: Why this tests seems to be testing nothing?

test/index.html
```diff
@@ -0,0 +1,250 @@
...
+      var done = assert.async();
+
+    });
```
FagnerMartinsBrack: Why this tests seems to be testing nothing?

test/qunit/qunit.js
```diff
@@ -0,0 +1 @@
+qunit-1.20.0.js
```
FagnerMartinsBrack: Why does this file contains only the filename?

test/qunit/qunit.css
```diff
@@ -0,0 +1 @@
+qunit-1.20.0.css
```
FagnerMartinsBrack: Why does this file contains only a  filename?

FagnerMartinsBrack:
Maybe we could split the tests into logical modules and leverage QUnit `before`/`after` APIs to setup requirements for each test (AKA [testDone](http://api.qunitjs.com/QUnit.testDone/) and [testStart](http://api.qunitjs.com/QUnit.testStart/)).

I am also wondering if it doesn't make more sense using more fluent testing APIs such as Mocha/chai instead of Qunit. Up to discussio.

test/index.html
henrikingo: I think I've tried to stay (mostly) within 80 character line length. In this case it wasn't necessary to have a line break, but for the longer explanations it is. I'm ok with not following such a rule though. You're correct impress.js certainly has lines longer than that.

test/index.html
henrikingo: I don't feel that one is significant at this point. The file is a work in progress, so wanted to leave a note for others to think about it, but it's not a blocker for going forward. To be clear: The tests work now. Ideally I would want to call impress().init(); inside a test, but it doesn't do that for now.

test/index.html
henrikingo: It's a placeholder for future tests. Given the discussion yesterday, I wanted to share what I have now. Oh, you're right, I should at least add a done(), so that the test finishes with 0 assertions.

test/qunit/qunit.js
henrikingo: It's a symlink to the filename with version numbers. I completely forgot there are Windows users in this world. Will remove the symlinks.

henrikingo:
Thanks for catching I had somehow added wrong email to .gitconfig. Fixed now.

I picked Qunit out of the discussion in #494. If someone wants to provide pull request with different framework, I won't be insulted :-) I just wanted to move things forward.

In hindsight, QUnit is awkward in 2 ways:
 - the idea is to manipulate html elements under a div, which is reset between tests. However, impress.js sets classes in body and wants to own the entire page. It still works, but this is not how QUnit developers intended it to be used. (hence need for cleanup yourself)
 - QUnit prints results on the screen, while impress.js wants to do animations on the entire screen as well. It works, but a bit messy. I believe if we run Qunit in karma or other test runner, we will get results printed in console, so this might be ok after all.

As for different modules: that's exactly the idea. This would be the first module testing core functionality. Keypress tests can go into this one, or could already be separate file. In any case more advanced animations need their own html file.

No comment yet on before/after api.

henrikingo:
I forgot to say: Thanks for reviewing!

test/index.html
FagnerMartinsBrack: I definitely agree with 80 characters limit, but since impress codebase does not seems to care about that, it would be better to keep consistency until we have setup a proper standard that everyone agrees upon.

test/index.html
FagnerMartinsBrack: I believe it's worth tracking down the root cause of everything and keep new technical debt at a minimum. Since you are the one currently inspecting the codebase, it is much easier for you to do this.

The reason why I suggest tracking and exposing the root cause of all problems via PR comments is that it allows us to understand better the current state of the code in order to think about better strategies to tackle those problems now or in the future.

test/index.html
FagnerMartinsBrack: > It's a placeholder for future tests.

Oh ok xD. Maybe we could prefix the PR with a `WIP:` in the title just like how GitLab does? This way we can mark a PR as a Work in Progress to help building some context before somebody start reviewing :)

@impress/mergers

test/qunit/qunit.js
FagnerMartinsBrack: Yeah, I use Windows for personal purposes, and Mac/Linux for Work.

FagnerMartinsBrack:
> I picked Qunit out of the discussion in #494

That was just my personal experience at the time, not that it really matters. We should probably focus on what is the best for the project anyway.

> QUnit is awkward in 2 ways

I agree QUnit is not that robust or fluent, I realized that after working with different testing frameworks over time such as Mocha/Jasmine integrated with Karma. But as I said earlier we need to judge which one is better to achieve the goal of taking this project forward.

> I forgot to say: Thanks for reviewing!

You're welcome. I hope the same for any PRs I send from now on :)

test/index.html
henrikingo: Working on your comments today.
I started removing the newlines, but for the QUnit.assert() functions I'm gonna keep the style as it was. The reason is that the last argument is the text that explains each assertion, kind of like a comment. It is of value to have those left aligned, rather than starting at an arbitrary position after all other arguments.

test/index.html
henrikingo: Still can't say I fully understand it, but at least I can call it inside a Qunit.test() now. (You'll see when I push later.) My theory for now is simply that some of the css classes aren't immediately readable in classList.contains()

test/index.html
FagnerMartinsBrack: I don't see the value of having the text aligned, after all it is a matter of personal opinion for code style. In my experience, code style discussions tend to take too many resources and start discussions that take forever, that's why it is a common practice in the community to go for consistency instead of discussing about it. That said, if this PR is landed and we choose to go for message alignment, then all tests from now on that uses QUnit should come with the same alignment for consistency. We could even document that style for consultation later.

...

After I wrote the paragraph above I started thinking: If we are having trouble with QUnit since the beginning, doesn't that count as an argument in favor of another more fluent test framework? An assertion like `impress.should.not.be( undefined )` make us not to worry about how to structure the test message and let the framework do it for us. I am definitely playing devil's advocate here, because I like QUnit simplicity.

What do you think, does it makes sense?

FagnerMartinsBrack: I wonder if we can use a fluent assertion library on top of QUnit, if that's possible, then all problems are solved!

test/index.html
henrikingo: Ah, now I got it. The "active" css class is set immediately in goto(), whereas "present" is set in impress:stepenter event handler. Moving the check for "present" to the test that checks event handlers.

test/index.html
henrikingo: Do you mean the beforeEach()/afterEach() functionality that is part of QUnit.module(). I don't think that applies here. Or it could, but depends on what the other tests do and what they need to clean up. For now, I'm assuming each test() block will do its own cleanup.

henrikingo: Ah, you mean testStart()/testDone(). Still, it's not obvious at this point that using that will be useful.

FagnerMartinsBrack: > Still, it's not obvious at this point that using that will be useful.

Those callbacks should be implemented for cleanup instead of an additional test. Tests should be atomic and not dependent of other tests. `testDone` will execute the desired routine and keep the tests atomic, since executing a single test will still keep running the cleanup, while, in the same situation, if implemented in a custom arbitrary test the cleanup will not run.

test/index.html
```diff
@@ -0,0 +1,250 @@
...
+                   event.target.id + " unset future css class." );
+        assert.ok( !event.target.classList.contains("past"),
+                   event.target.id + " unset past css class." );
+
+        if( nextStep[i].next ) {
+          assert.ok( nextStep[i].next(), nextStep[i].text );
+          i++;
+        } else {
+          // Remember to cleanup, since we're operating outside of qunit-fixture
+          root.removeEventListener( "impress:stepenter", registeredListener );
```
henrikingo: This didn't really remove the event listener. Fixing in next commit.

henrikingo:
@FagnerMartinsBrack : New commit should address all issues you raised (or intentionally not, as with the line breaks in assert() invokations).

test/helpers.js
```diff
@@ -0,0 +1,30 @@
+// Helper function to determine whether this browser is supported by 
+// impress.js or not. Copied from impress.js itself.
+var _impressSupported = function() {
```
FagnerMartinsBrack: Why the prefixed underscore?

test/helpers.js
```diff
@@ -0,0 +1,30 @@
...
+      return memory[ prop ];
+    };
+  })();
+
+  var ua = navigator.userAgent.toLowerCase();
+  return   ( pfx("perspective") !== null ) &&
+           ( document.body.classList ) &&
+           ( document.body.dataset ) &&
+           ( ua.search(/(iphone)|(ipod)|(android)/) === -1 );
+}
```
FagnerMartinsBrack: I believe we can assume that the browser we are testing on support all the features. If we start testing automatically later in browsers that do not have those features, then it makes sense creating a check like this.

test/index.html
```diff
@@ -0,0 +1,247 @@
...
+
+    QUnit.test( "Initialize Impress.js", function( assert ) {
+      assert.ok( impress, 
+                 "impress declared in global scope" );
+      assert.deepEqual( impress().init(), undefined,
+                        "impress().init() called." );
+      assert.deepEqual( impress().init(), undefined,
+                        "It's ok to call impress().init() a second time, it's a no-op." );
+      var notSupportedClass = document.body.classList.contains("impress-not-supported");
+      var yesSupportedClass = document.body.classList.contains("impress-supported");
```
FagnerMartinsBrack: Couldnt it be just `supportedClass`?

test/index.html
```diff
@@ -0,0 +1,247 @@
...
+        assert.equal( canvas.style.transitionTimingFunction,
+                      "ease-in-out",
+                      "canvas.style.transitionTimingFunction initialized correctly" );
+                      
+        assert.equal( document.documentElement.style.height,
+                      "100%",
+                      "documentElement.style.height is 100%" );
+        
+        // Steps initialization
+        var step1 = document.querySelector( "div#step-1" );
```
FagnerMartinsBrack: There seems to be inconsistencies for spaces inside parens

test/index.html
```diff
@@ -0,0 +1,247 @@
...
+    // TODO: more complex animations and check position, transitions, delays, etc...
+    // Those need to be separate html files, and there could be several of them.
+    QUnit.test( "Impress Core API", function( assert ) {
+      // CSS classes are themselves manipulated in event listeners. Wait a short while before checking, to avoid race:
+      var wait = 5; // milliseconds
+      // Remember, when this block of tests is done, we must disable our event listeners
+      var disableListeners = false;
+
+      var doneWithNext = assert.async();
+      var doneWithPrev = assert.async();
```
FagnerMartinsBrack: Is there a need to instantiate the `async` twice? When you call it the first time it will already indicate to the test suite should continue to run the remaining tests.

test/index.html
```diff
@@ -0,0 +1,247 @@
...
+        assert.ok( !event.target.classList.contains("active"),
+                   event.target.id + " unset active css class." );
+        assert.ok( !event.target.classList.contains("present"),
+                   event.target.id + " unset present css class." );
+        assert.ok( !event.target.classList.contains("future"),
+                   event.target.id + " unset future css class." );
+        assert.ok( event.target.classList.contains("past"),
+                   event.target.id + " set past css class." );
+        
+        if( i+2 == prevStep.length ) {
```
FagnerMartinsBrack: I really believe that each event should be tested in its own test block, then we can find a way to detach this kind of logic and magic number from the test assertions.

FagnerMartinsBrack:
Have you considered a fluent API for writing the assertions? This way we don't need to write the teste messages, it will be implicit from how the assertions are written.

test/helpers.js
henrikingo: I suppose it was just because this is internal helper. If you want to suggest a more proper way of encapsulating this function into some helper namespace, I will be listening :-)

test/helpers.js
henrikingo: That's not really why this is here. This is related to testing that some CSS classes were set by impress.js code:

      if ( !_impressSupported() ) {
        assert.ok( notSupportedClass,
                   "body.impress-not-supported class still there." );
        assert.ok( !yesSupportedClass,
                   "body.impress-supported class was NOT added." );
      } else {
        assert.ok( !notSupportedClass,
                   "body.impress-not-supported class was removed." );
        assert.ok( yesSupportedClass,
                   "body.impress-supported class was added." );
                   
If we just hard-coded assume that the test is run in a browser where impressSupported is true, then it defeats the purpose of the whole test. If the test ever fails, you wouldn't know whether it is because the browser check is failing, or because the code is setting css class incorrectly.

test/index.html
henrikingo: I guess I just wanted to emphasize the yes/no dichotomy.

test/index.html
henrikingo: True. Checking impress.js for guidance, it seems itself to be very inconsistent.

The logic in my code above is that short words go without spaces, like ´contains("impress-enabled")´ and longer (which is most) arguments are surrounded by space inside parenthesis.

My proposal is to leave this discussion for now, and later do an overhaul of both impress.js and this code where style is settled once and for all. (Preferably in the same pull request with adding a style checker.) This is not to say I'm opposed to comments about style in general.

test/index.html
henrikingo: The test is only done/passed when both doneWithNext() and doneWithPrev() have been called. They're called in different event listeners.

test/index.html
henrikingo: Even if it means copy pasting lines 130-150?

I agree with you so much that copy pasting might be easier to read, but I argue that this approach is easier to maintain. (And, as I've said before, your way test will run twice as long.)

henrikingo: Btw, if you wanted to focus on testing each event listener separately, it still leaves an open question: where would you check for the "active" css class? It is set inside the goto(), so not set inside either event listener. So should there be yet another test where that is checked separately?

This is just to underline my point: This test is not checking event triggers per se, rather it is testing everything related to moving from one step to next.

henrikingo:
Isn't that just another way of doing what I already do in QUnit? Wouldn't want to add more libraries for such a reason.

test/helpers.js
FagnerMartinsBrack: > If you want to suggest a more proper way of encapsulating this function into some helper namespace

The current code is not doing any encapsulation, it is exposed as a global variable with or without underscore. There is no difference between prefixing with underscore or not, in either case we need to take care to acknowledge the globals that are being used in testing and take care for not overriding it.

If one is worried about too many global variables, it is always possible to put the code into a closure and expose a single "helper" namespace that will contain all necessary utility classes as properties of the returned object.

Also, I would argue that this code is so similar to the one in the core that there is no problem in exposing it only for testing purposes to prevent code duplication, what matters is what is documented, not what is exposed (of course, it is necessary a proper public API documentation first).

test/helpers.js
FagnerMartinsBrack: Sure, that makes sense.

test/index.html
FagnerMartinsBrack: > My proposal is to leave this discussion for now, and later do an overhaul of both impress.js and this code where style is settled once and for all. (Preferably in the same pull request with adding a style checker.) This is not to say I'm opposed to comments about style in general.

Sure, we should hold style discussions when we have something that automatically checks it and when the codebase is consistent. I believe that could be one of the first tasks after the cleanup.

test/index.html
FagnerMartinsBrack: > The test is only done/passed when both doneWithNext() and doneWithPrev() have been called

Interesting, I have never used like that.

test/index.html
FagnerMartinsBrack: What is easier to read is also easier to maintain. Having something that is easier to maintain but hard to read seems to be a contradiction.

I understand your point about test execution time, the problem is that if you have too many premisses in a single test then it will be more dificult to create a deterministic assertion in the future when you still have to maintain the previous ones. Having a single atomic block of code that test a specific path turns out to be much more maintainable and scalable when developing regression tests.

What I suggest is that, instead of creating a test called "Impress core API", it should instead be created a "module" called "Impress core API" with tests that cover several situations separately. This way if one wants to run the tests that tests for only a specific event or animation flow, it is just clicking in the desired test in QUnit UI that it will go straight to it, without having to worry about the others. That is very useful for debugging purposes.

Below there are some examples:

**Found the failing test among others**

image

**Clicked on it to see why it is failing**

image

FagnerMartinsBrack: > Btw, if you wanted to focus on testing each event listener separately, it still leaves an open question: where would you check for the "active" css class? It is set inside the goto(), so not set inside either event listener. So should there be yet another test where that is checked separately?

Yes. An atomic test that checks a specific behavior is never a bad thing. It helps debugging and increase maintainability for not having too much logic inside a single block. Of course that can't be taken as a silver bullet, there are situations where, in order to test a specific behavior, you need several premises, but then that starts to smell like an architectural problem in the code that is being tested.

FagnerMartinsBrack:
QUnit is assertive, while a fluent API is behavioral. The benefits of a behavioral API is that you don't need to care how you are going to structure your assertions, you just declare your expectations and the API take care of presenting the failures automatically.

It's not like it is strictly necessary, just would be an interesting enhancement in case you want to look into that. We can do it later anyway.

test/helpers.js
henrikingo: Actually, I agree with your last suggestion. However, in this first commit, I didn't yet want to touch js/impress.js, since the religious belief is we must add tests before touching it. And I think it's cood to be overly conservative in these first steps. So I have copy pasted the lines I needed into this helpers file. In the future, I agree this code could be removed and instead we expose this functionality from impress.js itself.

FagnerMartinsBrack: >  In the future, I agree this code could be removed and instead we expose this functionality from impress.js itself.

Good point about not touching it for now. It makes perfect sense.

test/index.html
```diff
@@ -0,0 +1,247 @@
...
+    <div class="step" data-x="-600" data-y="0">Third slide</div>
+    <div class="step" id="fourth" data-x="-400" data-y="0">Fourth slide</div>
+  </div>
+
+  <div id="qunit"></div>
+
+  <script src="../js/impress.js"></script>
+  <script src="qunit/qunit.js"></script>
+  <script src="helpers.js"></script>
+  <script>
```
FagnerMartinsBrack: Shouldn't the tests be in a separate `*.js` file?

test/index.html
```diff
@@ -0,0 +1,247 @@
...
+    // In case tests didn't complete, we are left with a hash/anchor pointing somewhere. But we want to start from scratch:
+    window.location.hash = "";    
+
+    QUnit.test( "Initialize Impress.js", function( assert ) {
+      assert.ok( impress, 
+                 "impress declared in global scope" );
+      assert.deepEqual( impress().init(), undefined,
```
FagnerMartinsBrack: Why is `deepEqual` being used here? If the intent is to assert strict equality `strictEqual` should be used for that purpose. It's not that it is a big of a difference, is just that `deepEqual` tests if `expected` contains the same JSON-like structure of `actual`, while `strictEqual` do a simple strict equality check (`===`).

test/index.html
```diff
@@ -0,0 +1,247 @@
...
+  </div>
+
+  <div id="qunit"></div>
+
+  <script src="../js/impress.js"></script>
+  <script src="qunit/qunit.js"></script>
+  <script src="helpers.js"></script>
+  <script>
+    // In case tests didn't complete, we are left with a hash/anchor pointing somewhere. But we want to start from scratch:
+    window.location.hash = "";
```
FagnerMartinsBrack: Could this be declared inside of one of [QUnit callbacks](http://api.qunitjs.com/category/callbacks/)?

test/index.html
```diff
@@ -0,0 +1,247 @@
...
+        assert.ok( step1.classList.contains("active"),
+                   "Step 1 has active css class." );
+      }
+    });
+
+    // Note: Here we focus on testing the core functionality of moving between
+    // steps, the css classes set and unset, and events triggered.
+    // TODO: more complex animations and check position, transitions, delays, etc...
+    // Those need to be separate html files, and there could be several of them.
+    QUnit.test( "Impress Core API", function( assert ) {
```
FagnerMartinsBrack: This should definitely be a test module. In the future, a regression test related to the "Impress Core API" should be added to the "Impress Core API" module instead of being a standalone test. In QUnit AFAIK you can't create tests inside tests, only inside modules.

FagnerMartinsBrack:
Basically I see a few points of general improvement that I might have mentioned earlier or not:

* If there are comments in the code, then it means that the code cannot document itself. Is it possible to write the code in a way that it only documents what can't be seen by reading the code? If you want that I point out some examples in the code I will be glad to do it, it's just that I think it's pretty obvious that some comments are just redundant, maybe that's intended with plans to be removed later?

* Is it possible to split the JavaScript code in a separate `*.js` file? I say that because usually it helps for code editors to highlight the correct tokens in a file that only has JavaScript. Besides that, the separation might make it easier to focus in the behavior of the tests instead of how the document is structured (the html itself will be pretty small).

* Just out of curiosity, is it possible to run the second test `Impress Core API` separately? Ideally tests should be atomic in a way that they can be run separately without having dependencies in the other.

* About code style, as we probably already discussed, it is better to define and document some guidelines before this is landed (it is going to be done in the "CONTRIBUTING" part of https://github.com/impress/impress.js/issues/512). After we define it, we can adjust if necessary as a last step. Does that makes sense?

test/index.html
henrikingo: Are you referring to the TODO line? I realize now it's unclear (or incomplete, more correctly). My current thinking is that such tests are needed for proper coverage, but they would have to be their own html file. This is because tests operate on the div.step elements we've defined directly under body. So we can only have one impress.js presentation in one html file. The 4 steps in this html file are of course very simple, and allow for basic tests, like testing transitions. To test more advanced animations and positioning, another html file will be needed. But that will still be testing the Core API.

henrikingo:
> If there are comments in the code, then it means that the code cannot document itself. Is it possible to write the code in a way that it only documents what can't be seen by reading the code? If you want that I point out some examples in the code I will be glad to do it, it's just that I think it's pretty obvious that some comments are just redundant, maybe that's intended with plans to be removed later?

You're free to point out examples, but I think you'll learn that we have fundamental difference in philosophy here. In my taste, too much comments is better than too little. In my experience, most coders vastly overestimate the readability of their own code, when read by someone else. In particular, in open source we must take into account that code we write might be read by someone who never worked on this project before, and in the case of JavaScript, might be a person for whom JavaScript is not their strongest language (because they're really a backend dev, etc...) Yet, such persons should be treated as potential contributors too.

Also, it is intentional that code comments add redundancy. They're a statement of intent. You can then read the code to check whether the code does what the comment said. Without comments, you can read what the code does, but there's no reference point that will tell you whether that is what the code is supposed to do.

> Is it possible to split the JavaScript code in a separate *.js file? I say that because usually it helps for code editors to highlight the correct tokens in a file that only has JavaScript. Besides that, the separation might make it easier to focus in the behavior of the tests instead of how the document is structured (the html itself will be pretty small).

Sure. My editor handled this fine, but if it helps other, we should do that.

Actually, future tests (in other files) should have more elaborate html and css code, so this separation should definitively help in readability.

> Just out of curiosity, is it possible to run the second test Impress Core API separately? Ideally tests should be atomic in a way that they can be run separately without having dependencies on each other.

The first QUnit.test() calls `impress().init()`, so no. However, that is an expection and in general I agree with this principle. For example, my intent is to add tests for the keypress api into the same file. It should and would be possible to run the keypress tests without the core api tests. But both of them would depend on the first test calling impress().init().

As we cannot benefit from the automatic reset of the div#qunit-fixture element, it is actually very important to emphasize that each test cleans up after itself and tests don't depend on things done by previous test. (For example, it might become necessary to do things like `impress().goto(1)` between tests. QUnit seems to have an api to do such things between tests.)

> About code style, as we probably already discussed, it is better to define and document some guidelines before this is landed (it is going to be done in the "CONTRIBUTING" part of #512). After we define it, we can adjust if necessary as a last step. Does it makes sense holding this until the guidelines are defined by the team?

I agree on the first part: we shouldn't discuss style until there's a document that defines the guidelines.

I disagree on the second part: I think this should go in now. There are still lots of tests that need to be written, and adding QUnit will enable others to help writing tests. Also accepting any other code contributions is now blocked by not having tests. I of course hope we do agree on style guidelines reasonably quickly, but it would be backwards to block adding new code for a style discussion. Editing the code for style is a trivial task once we have agreed on the style guidelines, I will certainly quickly edit my own code if nobody else does it.

test/index.html
FagnerMartinsBrack: > Are you referring to the TODO line?

I was referring more to the fact that "Impress Core API" is declared as a `QUnit.test` instead of a `QUnit.module`. The name "Impress Core API" represents a set of tests instead of a single one. My suggestion would be something like this:

```javascript
QUnit.module( "Impress Core API" );
...
```

> This is because tests operate on the div.step elements we've defined directly under body. So we can only have one impress.js presentation in one html file.

Ahh, now I see. Having one test seems to be more a required need due to how impress works than an actual decision. In this case, to organise the tests accordingly inside each `test`, wouldn't be better run each one in a dedicated iframe that loads an external HTML file for each time the presentation should start from scratch? That would remove the worries about impress creating global data in the body element (or at least until we do it in a way where you can start more than one impress instance in the same page).

FagnerMartinsBrack: It could also be possible to create an utility that automates the iframe construction and destruction for us, just having to point an HTML file to be loaded before/after each test. Do QUnit have any plugin for this? Seems like an interesting use-case of a problem that somebody else could already have solved.

FagnerMartinsBrack:
> In my experience, most coders vastly overestimate the readability of their own code, when read by someone else

That usually happens when there is no code review (which is not the case for open-source projects maintained by more than one person) and when the developer is not aware of common principles of [idiomatic code](https://github.com/rwaldron/idiomatic.js/) and design patterns. It is definitely possible to create readable code as long as there is enough test coverage to provide safety or refactoring.

> You can then read the code to check whether the code does what the comment said

This is an unnecessary additional step to the whole understanding of the code. Also, if the code changes, developers tend to forget to update the comments. My previous statement is actually a fact, not opinion. You can see this happening in many popular open-source and closed-source projects. In jQuery, it is just watching the development for a few months that you will see this happening over and over again, mostly with browser support comments. Needless to say that we are talking about a lib that have more trust than any other library of its kind.

Besides that, I can quote a principle that is proven in the wild and used by the most successful companies in the world, one of which comes from the [agile manifesto](http://www.agilemanifesto.org/):

> Working software over comprehensive documentation

That principle was created due to the huge amount of evidence that existed and still exists among teams and individuals where internal documentation ends up not being maintained after a while, which always become outdated and obsolete.

Just to clarify that this do not happen with public API docs, because those docs represents the public contract to the consumer of the API, therefore acting as a single source of truth on how the API should behave. Basically if the public docs says something and the application doesn’t behave like that, then the application is the one that should be fixed, not the docs. If the public docs doesn't say something, then that is an undefined behavior. Needless to say that the documentation contract is pretty well understood by specifications such as semver.org, which has specific guidelines of versioning taking into account the public API.

> Without comments, you can read what the code does, but there's no reference point that will tell you whether that is what the code is supposed to do.

Yes, there is a reference point that will tell you what the code is supposed to do as stated in the previous comment, and that is the documentation of the public API.

> As we cannot benefit from the automatic reset of the div#qunit-fixture element

The idea that I commented earlier might solve this in a very unobtrusive way:
https://github.com/impress/impress.js/pull/510/files#r53293514

> it is actually very important to emphasize that each test cleans up after itself and tests don't depend on things done by previous test

I don’t understand this part. The tests are not cleaning themselves and the second one seems to depend in the previous one. Are you saying just to agree that the principle stands but that in this case it is impossible to do it because we can’t use quint-fixture?

> (For example, it might become necessary to do things like impress().goto(1) between tests. QUnit seems to have an api to do such things between tests.)

AFAIK QUnit only have APIs for running before/after all tests. Running before/after each module and Running before/after each test. AFAIK there is no API that runs in the middle of tests (only after or before each one).

> There are still lots of tests that need to be written, and adding QUnit will enable others to help writing tests. Also accepting any other code contributions is now blocked by not having tests.

These are very good arguments. I understand now why it makes sense doing this ASAP and I tend to agree with it after what you said.

FagnerMartinsBrack:
Just to make it clear that my arguments about comments in the code is not about comments that create value to the code, such as comments explaining something that the code doesn't (for any reason). It is specifically about redundancy, commenting what the code is explicitly doing.

[Here is an example](https://github.com/impress/impress.js/pull/510/files#diff-42a353e98c906af524c7d3d48feaf72aR1):

> // Helper function to determine whether this browser is supported by 
 > // impress.js or not. Copied from impress.js itself.
 > var _impressSupported = function() {

It is perfectly clear what the code means:

> Helper function

It is declared inside a helper.js file and is a function, therefore a helper function.

> to determine whether this browser is supported by impress.js or not

The name is actually `_impressSupported()`, which means that it checks if "impressSupported()"

> Copied from impress.js itself.

That is something unnecessary. A reader wouldn't need to know that. If he wants to find the discussion about why this code was added one can use `git blame` and eventually reach this PR via cross links that will contain the whole context of why that line was added in the first place. This is one of the reasons I love git and github so much :heart:

test/helpers.js
FagnerMartinsBrack: By the way, I was convinced about the point of not touching impress code, but I still don't see a need for the underscore as I commented above:

> The current code is not doing any encapsulation, it is exposed as a global variable with or without underscore. There is no difference between prefixing with underscore or not, in either case we need to take care to acknowledge the globals that are being used in testing and take care for not overriding it.

test/index.html
henrikingo: You mentioned QUnit.module() before, and I looked into it, but at this point I don't see that it would add any value. Maybe if there are more tests for the Core API in the future, then yes. But like you now understand, that might not actually happen.

I did briefly consider an iframe, but stayed with this approach now for simplicity. Also, I've been trying to keep as close as possible to how impress.js is typically used in real world, which is to own the complete window. Probably running in an iframe would be an equivalent test, but out of principle I want to test the "real thing" as closely as possible. I do think we should keep the iframe option in mind though, especially when more tests are added.

FagnerMartinsBrack: The good thing about doing this now (the iframe and the `QUnit.module` thing) is that we could prevent changing a lot of code in the future, but that is not the main reason. The main reason for using modules right now is that it forces us to create a better structure and understanding of separation os concerns on how the tests are structured and provide a good pattern to be followed from now on for the new tests.

I understand and agree about the arguments of simplicity though, but it might be worth a second look in the tradeoffs above.

> Probably running in an iframe would be an equivalent test, but out of principle I want to test the "real thing" as closely as possible.

If impress doesn't work inside an iframe as it would using the whole screen, then it should be reported as a bug. Impress should work in an iframe the same way it works in a regular window, I see no reason for it not to do so.

henrikingo: > The good thing about doing this now (the iframe and the QUnit.module thing) is that we could prevent changing a lot of code in the future, but that is not the main reason.

Also known as premature optimization :-) It's quite ok to change code once there's enough code to make it worthwhile.

I simply don't understand your pushing QUnit.module at this point. Currently there are exactly 2 QUnit.test blocks. If I put each of them in a separate module, how would that make any difference???

As for iframe, yes I know it works, I've seen it and developed code for impress-console. But "it should work" is not a thing you can say when we develop tests. We must start with testing how impress.js is primarily used. Then after that someone can do the extra work of proving that it also works 100% the same in an iframe.

Unless I'm mistaken, beyond this sub-thread there are no outstanding issues? Maybe it's time for me to squash the commits together into a single commit, and then merge this so we can proceed with adding more tests?

FagnerMartinsBrack: > Currently there are exactly 2 QUnit.test blocks. If I put each of them in a separate module, how would that make any difference???

It would leverage the QUnit API to provide separation of concerns. But that's ok, we can do this later.

> Unless I'm mistaken, beyond this sub-thread there are no outstanding issues?

There is only [this comment](https://github.com/impress/impress.js/pull/510#issuecomment-185647319) that points out an issue with comment redundancy. There are a few similar ones that just says what the code does explicitly.

henrikingo:
I will respond to your latest comment, as it contains a specific example:

Surely you agree, in principle, that each function at least should be preceded by a short description of what it does? I can promise you I will never submit a PR where that would not be the case.

More generally (to your previous comment):

Like I said in the beginning, we probably have a different philosophy here. I'm familiar with your thinking, and I have many friends who believe comments are unnecessary, too often outdated, etc. I consider them bad programmers. (Not because their code doesn't have comments, but because it has bugs.) The reason their comments are outdated is because they don't do their job well. They should have kept the comment up to date when code was changed, and if they didn't, it should have been caught in review. The programmers I respect write clear, commented code. (Also detailed commit messages, documentation, etc...)

Agile manifesto also says that we cannot collaborate on this project unless we sit in the same room so we can talk to each other. Within its context (enterprise software), it has done a lot of good, but personally I never felt I had much to learn from it. Open source movement has known how to write good code well before agile manifesto was written (and I'm old enough to talk about that).

Anyway, I think also this discussion falls within the style guidelines discussion. I will plead for my philosophy there, but I think we should commit this code now.

henrikingo:
Responding to this question separately:

> The idea that I commented earlier might solve this in a very unobtrusive way: https://github.com/impress/impress.js/pull/510/files#r53293514

Yes. I think it is a good idea. It's also enough work that it can be done in its own task, at a later time.

> I don’t understand this part. The tests are not cleaning themselves and the second one seems to depend in the previous one. Are you saying just to agree that the principle stands but that in this case it is impossible to do it because we can’t use quint-fixture?

All tests depend on `impress().init()` being called. There's no point in doing that repeatedly. Apart from that, tests must clean up after themselves. For example second test must remove the event listeners that it added, which it does.

> AFAIK QUnit only have APIs for running before/after all tests. Running before/after each module and Running before/after each test. AFAIK there is no API that runs in the middle of tests (only after or before each one).

Yes. I don't see what problem you're discussing here?

FagnerMartinsBrack:
> Yes. I don't see what problem you're discussing here?

You said "it might become necessary to do things like impress().goto(1) between tests. QUnit seems to have an api to do such things between tests.", then I just pointed out that it seems QUnit doesn't have an API to do things **between** tests, just **before**/**after** each one, which is not the same.

> Surely you agree, in principle, that each function at least should be preceded by a short description of what it does?

Let's talk about this specific case. If you are talking about internal **test** functions whose comments does not generate a public API, then definitely not, not me nor anyone, there is no default rule or principle on that matter, it all depends upon the context. For example, there is not a single popular open-source project that is trusted by many companies and maintained by more than one person that always documents **internal test** functions by default. In this case, a function preceded by a description of what it does is an evident code smell, and is only used when the code cannot be changed by other reasons (which is clearly not the case for this PR or the examples I mentioned [in the other comment](https://github.com/impress/impress.js/pull/510/files#r53445859)) or does not make explicit what it does.

> Agile manifesto also says that we cannot collaborate on this project unless we sit in the same room so we can talk to each other.

I never said anything about using any agile methodology here and there is no point in discussing about that. I never use agile in open-source projects for obvious reasons. This is probably an unintentional [straw man](https://en.wikipedia.org/wiki/Straw_man) that was created in this discussion due to an incorrect interpretation of what I meant. What I said is that one of the most evident principles in software development that is also used in one of the most widely used software development methodologies already predicts that documentation should not be abused: "Working software over comprehensive documentation". Creating redundant documentation is totally abusive when you already can use the tests to document how the application is supposed to behave.

> Anyway, I think also this discussion falls within the style guidelines discussion.

I definitely disagree with that. Writing useful code has nothing to do about code style, it can't even be verified by a tool like JSCS. It is about writing useful code instead of writing tons of comments to workaround the effort of creating good internal abstractions or comments that add value to the tests. There are very good comments in this PR, I just pointed out some obvious redundant exceptions.

Again, I am not against comments, I am against redundant comments in tests. Impress source code have a very good set of comments that annotate very well what it is supposed to do, sometimes even redudant for more experienced devs, but that is intended because the file is supposed to be used as a learning resource for beginners.

------

Since it doesn't seems we will get into an agreement on that, I will summon someone else from @impress/mergers or @bartaz to put and end into this discussion so we can get this landed once and for all.

To get a little bit of context and TL;DR for the ones that are coming here without having to read the discussion since the beginning, the discussion here is basically about useless comments in test files.

Take for example [this](https://github.com/henrikingo/impress.js/blob/f6349031631e859104555bd8fc078eb38508a38c/test/helpers.js#L1), [this](https://github.com/henrikingo/impress.js/blob/f6349031631e859104555bd8fc078eb38508a38c/test/index.html#L175) and [this](https://github.com/henrikingo/impress.js/blob/f6349031631e859104555bd8fc078eb38508a38c/test/index.html#L194). Should those kind of redundant comments be kept in test files, or we just document what cannot be infered from the code?

- [ ] Keep the redundant comments in test files
- [ ] Remove comments from test files when the code is explicit on what it does

When the majority of the team decides what it should be, I will agree with the decision and after that this PR is good to go :+1:

bcipolli:
>Take for example this, this and this. Should those kind of redundant comments be kept in test files, or we just document what cannot be infered from the code?

Fascinating discussion. Thanks @henrikingo for your coding and documentation efforts, very impressive to try and document to this level. And thanks @FagnerMartinsBrack for your review efforts.

I didn't expect to, but I actually agree with @FagnerMartinsBrack. For sure, there are many helpful comments in the code. At the same time, a few of the comments do seem unnecessary, and I think the code would be slightly more maintainable without them. Either way the decision goes, I don't think it's a big deal for this instance, however--it's very few lines of comments for functions/blocks of code that seems unlikely to change.

I do think the comment specifying code was copied from the main file is useful, perhaps should be marked with a `TODO`--to abstract the code out from both places at some point. No guarantee any of us will be around then to remember to do that, or to make sure the two versions are still in sync.

Anyway, just chiming in to let you know I'm here and interested to help! :)

FagnerMartinsBrack:
@bcipolli Thanks for the comment!

Could you please point out the lines whose comments you think should be removed from the PR, if any? Even though we can remove or improve anything later when we add more tests, it is still better to document the decision so that we can revisit later and not leave it one sided.

Thanks!

bartaz:
Hi guys, thanks for all the work you've put in here: @henrikingo for starting and writing tests and @FagnerMartinsBrack for very detailed reviews.

I just want to join the discussion here, because it seems that this review here is taking quite long and doesn't seem to move forward to finally merging it. So let me point some things to hopefully clean up the test code and have it included as a base in impress.

First of all I treat this PR as just a beginning - setting up of QUnit and some basic tests for impress.js, so it can be extended in the future and anyone willing to contribute will know where and how to put new tests.

From this point of view I would like to see the tests in separate JS file and not included in HTML as it is now.

My second biggest issue is with "Impress Code API" test, which is very big (with lots of internal functions), it's not clear for me what it tests, and it's quite hard to read. It seems to me that the sequence of the actions tested is coded in an array, and it's then performed and asserted, but the amount of nested functions makes it quite hard to understand (even for someone who knows impress.js). The fact that for example `assertStepEnter` and `assertStepLeave` use the value of `i` that is defined and changed outside of their scope is really confusing for me.

Of course I understand that this complexity is partly caused by impress.js itself and the fact that it's features are quite hard to test (in terms of unit testing) right now.

@henrikingo Would it be possible to change the "Core API" test from this 'loop-based' sequence into more coded sequence of actions and checks? This I think would make it more readable. Something like:

```javascript
QUnit.test( "Impress Core API", function( assert ) {
    // ... writing kinda pseudo-code here
    root.addEventListener("impress:stepenter", stepEnterMocked);
    root.addEventListener("impress:stepleave", stepLeaveMocked);
    impress().goto(2);
    assert.equals(currentStep, step2);
    assert.ok(currentStep.classList.classList.contains("active");
    // .. more asserts here
    impress().next();
   assert.equals(currentStep, step3);
   assert.equals(prevStep, step2);
   // .. etc
```

Where `stepEnterMocked` `stepLeaveMocked` will be some simple handlers that would just store target of event in `currentStep`/`prevStep`. I know it may be more complicated because of async nature of events but maybe it would be possible?

I'm also not sure about the need of `setTimeout` - maybe I'm missing something. It's true that impress.js internally uses events to set classes on elements, but these events are registered internally by impress.js when its initialized, so if in test we register an event (even on the same element) it should be called after internal handlers from impress.js. So when any external handler is called elements should already have the classes assigned to it. Is it not the case?

These are the things I would focus on in this PR to finish and include it.
Any discussion about code style or comments I would leave for now. We don't have any official code style (and I myself don't use the same style I used when writing impress.js anymore) so this can be updated later. And in terms of comments it may be topic for another discussion, but I prefer to have them, especially when they tell why (or how) something is done - not what is done (which should be self explanatory from the code). 

Sorry for this long comment, I'll try to keep an eye on this PR because it seems valuable for future changes to impress.js.

Thanks everyone for all the work you done here so far and let me know what you think about my suggestions here.

FagnerMartinsBrack:
:point_up: @bartaz I agree with everything you said, including about being declarative and the observation on the comments (which is to document only what the code can't tell for itself).

henrikingo:
Hi @bartaz 

Thanks for joining, I think your point of view was really needed.

> From this point of view I would like to see the tests in separate JS file and not included in HTML as it is now.

This is ok. I already have done that in my own fork, but didn't want to push more into this branch yet.

> I'm also not sure about the need of setTimeout - maybe I'm missing something. It's true that impress.js internally uses events to set classes on elements, but these events are registered internally by impress.js when its initialized, so if in test we register an event (even on the same element) it should be called after internal handlers from impress.js. So when any external handler is called elements should already have the classes assigned to it. Is it not the case?

Most of the things impress.js does after goto() is done in the [event listener] (https://github.com/impress/impress.js/blob/master/js/impress.js#L586). So when in test code we add an event listener, as far as I understand, those are equals and execute in parallel. You can test this by registering the event handler without a setTimeout wrapper:

    root.addEventListener( "impress:stepenter", assertStepEnter );
    root.addEventListener( "impress:stepleave", assertStepLeave );

With the above you will either get test failures (because not all CSS classes are set yet) or the test will hang (I believe because test framework called the next impress().goto() too fast).

So the setTimeout is unfortunately needed to ensure that impress.js event listener has finished before the test code checks for its results. This could of course be changed, but we would have to change impress.js code to do so, and now the primary goal was to add tests before any impress.js code is touched.

Note that apart from the above explanation, the setTimeout of course doesn't add any other value to the testing. It's true that it is an unfortunate additional layer of indirection.

> Would it be possible to change the "Core API" test from this 'loop-based' sequence into more coded sequence of actions and checks?
> ...
> Where stepEnterMocked stepLeaveMocked will be some simple handlers that would just store target of event in currentStep/prevStep. I know it may be more complicated because of async nature of events but maybe it would be possible?

It is of course possible, and if even you have a hard time reading the current code, then maybe we need to think about it. But it is certainly not going to be as elegant as your above pseudo code suggests.

First of all, all the asserts of course need to be done inside the event handler. Otherwise they will execute immediately, before even the first ´next()´ has completed and triggered the first event(s).

From that observation, the second problem is that if I'm not allowed to use a loop structure to make the event handlers parametrized, I think the only way to do this would be for each event handler to end with removing itself, and adding the next event handler.

So, just to give a short example:

    QUnit.test( "Impress Core API", function( assert ) {
      var eventHandler1_enter = function(event) {
         assert.ok(...)
         assert.ok(...)
         ...
         impress().next()
         assert.ok( /* the things currently in assertImmediately() */ )
         ...
         root.removeEventListener( eventHandler1_enter )
         root.addEventListener( eventHandler2_enter )
       }

    var eventHandler2_enter = function(event) {
       ...

(Note that above the wrapper calling `setTimeout()` is omitted for simplicity, but would of course have to be added back for the real code. And if we want to be consistent in this direction, there would now be a different `setTimeout()` wrapper for each step as well?)

There would be 7 such event handlers for `stepenter`, another 7 for `stepleave`. Each containing 6-8 assertions. So in total by "unwinding" the current loop approach, we would be adding about 110 lines of code. Note that this then repeats if we have to use the same style for testing the key and mouse navigation (which has 12 transitions instead of 7!), and probably more future tests as well. So just for this js file we're talking about 200+ additional lines of code.

Finally, I think both you and @FagnerMartinsBrack have struggled with the fact that we're checking both `stepenter` and `stepleave` in parallel, and can call `impress().next()` only after both code paths are finished. This is why the `readyCount` and `readyForNext()` functions are used to start the next step. So to simplify this part we would have to (as already suggested by @FagnerMartinsBrack) separate the tests checking for `stepenter` and `stepleave` into their own `QUnit.test()` blocks. This will actually not add much more lines of code (maybe 10-12 or so) but will double the execution time of the tests. But that is acceptable IMO if we really want to go this way. 

Much bigger problem is code maintainability, when same code is copy pasted 7 times. I can say for sure that it would not have been feasible to develop the tests in the style we're now discussing, since every simple change will then have to be done 14 different times (or 24, for the key/mouse navigation). But of course now that the tests are done and working, I can unwind them into a longer, more readable format, with the expense of making future maintenance of the tests more laborous.

My compromise suggestion here would be to do the last part: We can separate the `stepenter` and `stepleave` tests into their own `QUnit.test()` blocks, and get rid of the `readyCount / readyForNext()` logic, but keep the loop structure (to avoid serious code bloat). At least I can easily do this first with little effort, then you can judge if code is any more readable with such a change or if more unwinding is needed.

FagnerMartinsBrack:
Even if the suite takes more time, we can reduce the time later, for example, by testing  without waiting for transitions or creating better preparations before/after each test.

There seems to be a lot of complexity in the process that maybe could be simplified a lot if we talk about each problem in isolation:

> Much bigger problem is code maintainability, when same code is copy pasted 7 times.

Which code will be pasted 7 times? Can you give us an example? What does that code tests? And why there will be a problem with code maintainability?

> I can say for sure that it would not have been feasible to develop the tests in the style we're now discussing, since every simple change will then have to be done 14 different times (or 24, for the key/mouse navigation).

Can you please give an example of a simple change that would make it necessary to be done 14 different times? I mean, we should't try to aim for 100% coverage right in the beginning, if that's what you are talking about.

henrikingo:
Hi @FagnerMartinsBrack, and thanks for asking

> Which code will be pasted 7 times? Can you give us an example? What does that code tests? And why there will be a problem with code maintainability?

So, as is probably clear by now, we want to test different transitions: `goto(), prev(), ǹext()` etc... After each transition, we verify several css and other properties, the same assertions each time. So if I cannot use a loop, naturally the contents of the current loop have to be copypasted into a sequence instead. Specifically, we are talking about these lines from `assertStepEnter()`:

    assert.equal( event.target, sequence[i].entered,
                  event.target.id + " triggered impress:stepenter event." );
    assert.ok( event.target.classList.contains("present"),
               event.target.id + " set present css class." );
    assert.ok( !event.target.classList.contains("future"),
               event.target.id + " unset future css class." );
    assert.ok( !event.target.classList.contains("past"),
               event.target.id + " unset past css class." );
    assert.equal( "#/"+event.target.id, window.location.hash,
                  "Hash is " + "#/"+event.target.id );

and also these, from `assertImmediately()`:

    assert.ok( sequence[i].entered.classList.contains("active"),
               sequence[i].entered.id + " set active css class." );
    assert.ok( !sequence[i].left.classList.contains("active"),
               sequence[i].left.id + " unset active css class." );      

Then the same has to be done for all assertions in `assertStepLeave()` as well.

> Can you please give an example of a simple change that would make it necessary to be done 14 different times?

Say that for some reason, the active step would no longer have a `present` css class, but we instead want to call it `current`. This would be a simple change in impress.js. But for these tests you would have to repeat that change for every time you copy pasted those assertions.

(Obviously, while developing tests, it would be impossible to have to do all the try and error with such an un-looped code, since the whole process of writing tests becomes 14 times slower for every edit.)

FagnerMartinsBrack:
I guess the problem is much clearer now, so here are my insights.

> Say that for some reason, the active step would no longer have a present css class, but we instead want to call it current. This would be a simple change in impress.js. But for these tests you would have to repeat that change for every time you copy pasted those assertions.

Sure, and in this specific example There is no problem with that. All tests that rely on the `present` class will need to be changed because they will fail anyway. Increase the abstraction just to change in a single place for a test has no value to prevent this sort of things, I will explain why.

Let's take a look in the benefits of abstracting code and using DRY in the source code:

1. You reduce the amount of bytes from the codebase, which is definitely a must in order to reduce the library size for the end-user.
2. If we change a `.present` somewhere in the source, we might forget to do that in another part and the tests might not get this bug. This is one of the the main reason why [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) exists. The 100% coverage is pretty hard, usually we start testing a single common path, assuming that the code is properly DRY and reused in many parts of the system, then one common test has a higher probability to cover more than a single path of the program execution due to this assumed reusability inside (which can easily be spotted by code review, a static testing pattern).

Now, those benefits apply to testing code?

1. In testing code we don't care about bytes, because that will not be downloaded by the end-user.
2. If we change `.present` to `.current` in the **source code**, then we need to take extra care in code review to know that it will be changed everywhere it is being used. In the tests though, it will just break all tests that are checking for `.present`, no need for extra effort because the tests are declarative. If a test does not break if the behavior changes, then the test is non deterministic and therefore is useless and should be removed.

Another benefit in changing the test in the places it is being directly asserted is that the reviewer will know exactly, by the diff, which tests are being affected by the change, instead of having to understand the whole abstraction and complex logic first. This might expose other things, such as useless tests.

I remember a quote I read once, don't remember the author which was something like this:

> You can make good tests become faster, but you can't make faster tests become good

Basically it means that we can fix problems in the future if we start in a simple way, incrementing the value of the code over time. This way, we can visualize the need for complexity and discuss as a team later.

My suggestion would be something like this (pseudo code):

```javascript
QUnit.module( "Core API", {
  before: function() {
    // Reset the whole impress.js state
  }
});

QUnit.test( "Initial state", function() {
  // Given 2 steps
  test.assert( /* step 1 has active/present classes */ );
  test.assert( /* step 1 triggered the stepenter event */ );
  test.assert( /* step 2 has future class */ );
  test.assert( /* hash is correct */ );
});

QUnit.test( "Left/Right move", function( test ) {
  // Given 3 steps
  // (The initial state is already tested)
  impress.next();
  test.assert( /* step 2 has active/present classes */ );
  test.assert( /* step 2 triggered the stepenter event */ );
  test.assert( /* step 1 has past class */ );
  test.assert( /* step 3 has future class */ );
  test.assert( /* hash is correct */ );
  impress.prev();
  test.assert( /* step 1 has present/active classes */ );
  test.assert( /* step 1 triggered the stepenter event */ );
  test.assert( /* step 2 has future class */ );
  test.assert( /* hash is correct */ );
});
```

Of course we need to account for async stuff and other requirements for the code to work, but always keep in mind to create declarative tests so that it serves also as a documentation of what is supported for future and existing maintainers. If it breaks, then we need to change it exactly in the specific assertion, so we know why it broke without having to understand the underlying logic behind it.

henrikingo:
I'm sorry, I don't quite understand what you're arguing for or against.

It's an undisputable fact that to change something in 14 different places (let alone 24, etc...) takes 14 times more time than changing it in one place. I don't know why you insist on ignoring this fact, but since I am the person actually writing this code, I happen to care about it quite a lot.

bartaz:
That's true - repetition (especially many times) is never a good sign. I don't want to be a devils advocate, but I would argue that 14 repetitions don't always mean 14 times more time, because short but complicated code may take similar amount of time to understand and do this single required change.
And also if there is so much repetition in tests, maybe we are testing the same thing couple of times unnecessarily?

We need to find a balance between not repeating too much while keeping it simple enough to understand and change. The hardest part of programming I guess ;)

But the truth is (and I'm guilty too looking at my previous and this comment) that we are in my opinion writing too much text in this discussion not making much progress.

Clearly the "Core API" test part is a biggest point in this discussion right now and source of confusion.
@henrikingo, I understand why you used loop approach to avoid duplication but in the current state it's quite hard to understand what this tests does (what it tests) and how to change or add anything to it.
To make test suite to be useful and used by other people (for example to require PRs having their tests) we need to make it simple - as far as it is possible with qunit and impress constraints. It's hard as far as we see in this discussion.

I would like to have tests in impress.js, it's great @henrikingo that you managed to set them up and got them running in quite unfriendly asynchronous and animation based impress.js environment. First step - the base - is already here. I would suggest to - at least for now - leave the confusing "Core API" part and merge the tests without it. Or reduce it to single test without loop (just check if calling `goto` changes step and classes on it). This will give us the base to add another tests later and work on any repetitions if necessary step by step.

FagnerMartinsBrack:
> It's an undisputable fact that to change something in 14 different places (let alone 24, etc...) takes 14 times more time than changing it in one place. I don't know why you insist on ignoring this fact, but since I am the person actually writing this code, I happen to care about it quite a lot.

Please, do not start to act defensively assuming that I am ignoring principles or facts, I acknowledge it perfectly. I am here to help, just like you :+1: 

The fact is that if there is a need to change the test in 14 different places then something is wrong, in the example I gave above there are only 3 assertions in which the `.present` class will have to be changed in order to get a very good initial coverage, which is reasonable. After testing the correct inclusion of the `.present` class once (or for each API, maybe), then there is no need to test it again all the time and we can focus in testing other stuff. Also, by arguing that the `.present` class will be changed we are creating complexity to defend ourselves against a speculative future: we don't know if the class will be changed, and if it does we can act accordingly in the future do handle any problems that come with that.

I guess the problem is that this PR does more than one thing, and that makes it harder to understand because there are a lots of assumptions in the current code. Ideally, a Pull Request should be atomic and do one thing, and one thing only. We should start simple, with little code as possible, and then increase complexity over time, if necessary.

**So, here is my suggestion, please tell me what you guys think about it:**

Start with the QUnit structure and the JavaScript code separate in a different file (which is not a speculative future, we will need that eventually), and a single atomic test (not module), that can be run alone or with the whole suite.

For example: Test only the `.next()` API. Assert that it adds the past/present/future and active classes in a single PR. After it is merged, we can create a new PR that adds a test for something else, so we can focus in only one thing at a time.

@henrikingo @bartaz Does that makes sense? I guess this will make the communication much better and help to land things faster :)

henrikingo:
New push.

Forgot to say in commit message that I renamed from index.html to core_tests.html, because when I put the tests into a separate .js file I didn't want to call it index.js. (as this could lead people it has special meaning for example within an npm context.) Ie. the new javascript file is then called core_tests.js.

FagnerMartinsBrack:
@henrikingo Thanks for the update!

I have taken some time to actually run the code and check it deeply. The previous reviews were only static, based on the assumption that everything works as expected. I never actually run the code.

So, given this deep review my insights are below:

FagnerMartinsBrack:
The second test doesn't run alone. Is it intended? Can we do it in a way that each test is atomic and can be run separately?

screenshot 2016-03-10 21 46 44

test/core_tests.js
```diff
@@ -0,0 +1,138 @@
...
+  if ( !_impressSupported() ) {
+    assert.ok( notSupportedClass,
+               "body.impress-not-supported class still there." );
+    assert.ok( !yesSupportedClass,
+               "body.impress-supported class was NOT added." );
+  } else {
+    assert.ok( !notSupportedClass,
+               "body.impress-not-supported class was removed." );
+    assert.ok( yesSupportedClass,
+               "body.impress-supported class was added." );
```
FagnerMartinsBrack: Now thinking again about this, does it makes sense testing that the `.not-supported` class is being added to the body element and vice-versa? I mean, we only run this test manually for now, and probably in evergreen browsers that support impress.js in a decent stability (Chrome for example).

Ideally this sort of test should be assertive on what it expects (or do not expect) to happen without caring about the environment it is running. If necessary, we should simulate an environment to make impress.js trigger the secondary condition to test if the `.not supported` is being properly executed, otherwise it is not going to run in every run and therefore we are not going to catch it (only if we run in a different browser that specifically doesn't support impress.js).

I suggest just asserting that it adds the class if it is supported and then later find a way (stubbing or mocking some data) to assert that the `.not-supported` class was added correctly.

test/core_tests.js
```diff
@@ -0,0 +1,138 @@
...
+  // In case tests didn't complete, we are left with a hash/anchor pointing somewhere. But we want to start from scratch:
+  window.location.hash = "";    
+});
+
+QUnit.test( "Initialize Impress.js", function( assert ) {
+  assert.ok( impress, 
+             "impress declared in global scope" );
+  assert.strictEqual( impress().init(), undefined,
+                    "impress().init() called." );
+  assert.strictEqual( impress().init(), undefined,
```
FagnerMartinsBrack: What is the value on testing multiple calls of impress.js API if we are not objectively determining if the behavior was actually noop or not (like checking some global variable or something like that?). This seems to be just a smoke test to ensure that multiple impress.js calls don't trigger an error, is my assumption correct?

FagnerMartinsBrack: If my assumption above is correct, then we don't need to assert the API return, which is void and therefore shouldn't be relied upon. Testing a feature marks it as supported, which can cause confusion. We should have green light to change the `impress.init()` return in the future, if necessary.

test/core_tests.js
```diff
@@ -0,0 +1,138 @@
...
+    // weren't supported.
+    assert.ok( !document.body.classList.contains("impress-disabled"),
+               "body.impress-disabled is removed." );
+    assert.ok( document.body.classList.contains("impress-enabled"),
+               "body.impress-enabled is added." );
+    
+    var canvas = document.querySelector( "div#impress > div" );
+    assert.ok( !canvas.classList.contains("step") && canvas.id === "",
+               "Additional 'canvas' div inserted between div#impress root and steps." );
+    assert.equal( canvas.style.transform,
```
FagnerMartinsBrack: It is a common pattern to declare `expected`/`actual` variables to preven [this kind of problem](https://github.com/jquery/jquery/commit/d3d8d9751f3d14a545b26cf820dc1f51896a7b50) in the future, that will screw with QUnit reports.

FagnerMartinsBrack: Also, it is a good practice to use strict equality when comparing two data types to prevents traps like this one:

```javascript
assert.equal( null, undefined ); // Passes, data is equal?
```

The solution to prevent this trap in the future is to use `strictEqual` instead:

```javascript
assert.strictEqual( null, undefined );// fails, data is different!
```

test/core_tests.js
```diff
@@ -0,0 +1,138 @@
...
+// TODO: more complex animations and check position, transitions, delays, etc...
+// Those need to be separate html files, and there could be several of them.
+//
+// For now this just contains a single transition in order to get QUnit committed.
+// Complete test coverage needs to be added later.
+QUnit.test( "Impress Core API - Work in progress", function( assert ) {
+  // impress.js itself uses event listeners to manipulate most CSS classes. 
+  // Wait a short while before checking, to avoid race. 
+  // (See assertStepEnterWrapper and assertStepLeaveWrapper.)
+  var wait = 5; // milliseconds
```
FagnerMartinsBrack: I managed to remove the need for this timeout with the following code:

```javascript
var root = document.querySelector( "div#impress" );
var step2 = document.querySelector( "div#step-2" );
var done = assert.async();
root.addEventListener( "impress:stepenter",  function( event ) {
  assert.strictEqual( event.target, step2, "Triggered event for the second step" );
  done();
});
impress().next();
```

In this case it triggers the test correctly without having to use `setTimeout, is there anything I am missing here?

FagnerMartinsBrack:
I haven't finished taking a deep look into the implementation. I will continue tomorrow :+1:

FagnerMartinsBrack:
I have created a Pull Request on top of this PR with some suggestions: https://github.com/henrikingo/impress.js/pull/1

FagnerMartinsBrack:
@henrikingo Just pinging to know if you are still tracking this.

Let me know if there is anything I can do to help or make any of my suggestions above clearer.
I ask this because #535 seems to be sort of dependent on this PR, since there is more than just white-space changes and a test is always welcome.

FagnerMartinsBrack:
@henrikingo @bartaz Pinging one more time. If there is no comment I will push https://github.com/henrikingo/impress.js/pull/1 through.

PR #530, opened 2016-03-07
fixing scale via data attributes in chrome

ktabors:


FagnerMartinsBrack:
@ktabors I will assume this was working before? Did chrome changed anything that broke this? Can you share a link explaining the issue?

Looking forward on your feedback.

ktabors:
This was my first time using the project. I didn't look into why this was a problem in Chrome, maybe I'm using it wrong. In Chrome dev tools I saw was the variable names in the object didn't have any capitals. From the html 'data-maxScale="1"'. I didn't check my solution with other browsers. Maybe there should be a check to see if the variable exists with or without capitals?

image

FagnerMartinsBrack:
> I didn't check my solution with other browsers

Please, can you do it? We have built a [contributing guidelines](https://github.com/impress/impress.js/blob/f9c41968eb0f06b1c20e3f631e1d7e5a464baf55/CONTRIBUTING.md#pull-requests) for guidance.

---

Anyway, the [spec states that](https://html.spec.whatwg.org/multipage/dom.html#dom-dataset):

> The algorithm for getting the list of name-value pairs:
> For each content attribute on the element whose first five characters are the string "data-" and whose remaining characters (if any) **do not include any uppercase ASCII letters**, in the order that those attributes are listed in the element's attribute list, add a name-value pair to list whose name is the attribute's name with the first five characters removed and whose value is the attribute's value.

Here is a test case: https://jsfiddle.net/k1L8sua1/

Conclusion: You should not use `data-maxScale` in the data attributes, but instead `data-max-scale`. Data attributes ignore uppercase characters.

See this fiddle for an example: https://jsfiddle.net/k1L8sua1/1/

FagnerMartinsBrack:
Closing as from the comment above.

PR #388, opened 2014-04-28
hide cursor in fullscreenafter a configured delay without movement

sclausen:
For making presentations better, it would be great, if the mouse cursor hides after going to fullscreen.
This commit implements this behavior.

PR #536, opened 2016-03-19
added option to hide cursor in fullscreen mode

sclausen:
making a new pull request as a replacement for #388 based on master instead of dev.

### DIFF
--- a/index.html
+++ b/index.html
@@ -144,7 +144,7 @@
     right around the corner of this tag ;)
     
 -->
-<div id="impress">
+<div id="impress" data-hide-mouse-after="3000">
 
     <!--
         
--- a/js/impress.js
+++ b/js/impress.js
@@ -397,6 +397,33 @@
                 scale:     1
             };
             
+            // Hide the mouse cursor in fullscreenmode after the configured
+            // amount of milliseconds without movement
+            if(rootData.hideMouseAfter) {
+                var mouseMoveTimer = null,
+                isCursorVisible = true;
+                
+                var hideCursor = function() {
+                    //isFullscreen
+                    if(!window.screenTop && !window.screenY){
+                        mouseMoveTimer = null;
+                        document.body.style.cursor = "none";
+                        isCursorVisible = false;
+                    }
+                };
+                
+                document.onmousemove = function() {
+                    if (mouseMoveTimer) {
+                        window.clearTimeout(mouseMoveTimer);
+                    }
+                    if (!isCursorVisible) {
+                        document.body.style.cursor = "default";
+                        isCursorVisible = true;
+                    }
+                    mouseMoveTimer = window.setTimeout(hideCursor, rootData.hideMouseAfter);
+                };
+            }
+            
             initialized = true;
             
             triggerEvent(root, "impress:init", { api: roots[ "impress-root-" + rootId ] });


index.html
```diff
@@ -144,7 +144,7 @@
     right around the corner of this tag ;)
     
 -->
-<div id="impress">
+<div id="impress" data-hide-mouse-after="3000">
```
henrikingo: Hi sclausen

Good idea and easy to implement.

IMO this feature needn't be optional. (Not adding an option benefits the goal of simplicity.) I don't think anyone will miss the pointer if we just hide it always after 3 seconds of inactivity. (I can see that someone would want to leave the mouse pointer pointing at something on a slide. But that should be rare. You'd usually move the pointer to point at something, but that would still work. Also, it is more common to use laser pointers, etc.)

js/impress.js
```diff
@@ -397,6 +397,33 @@
                 scale:     1
             };
             
+            // Hide the mouse cursor in fullscreenmode after the configured
```
henrikingo: We are moving to a plugin architecture. So instead of a patch against impress.js, you should write this as a plugin. See https://github.com/henrikingo/impress.js/blob/myfork/src/plugins/README.md for a detailed explanation. If you want to write a PR against that repo, you're welcome to do so.

Alternatively, if you want to do a PR against the impress/impress.js repo in its current state, you could still write this as a "plugin" By adding a separate `(function(window, document){...})(window, document);` at the end of the file. You can take a queue from the keypress navigation at the end of the file, which is in fact such a plugin (even if it's not in a separate file).

js/impress.js
```diff
@@ -397,6 +397,33 @@
...
+                var hideCursor = function() {
+                    //isFullscreen
+                    if(!window.screenTop && !window.screenY){
+                        mouseMoveTimer = null;
+                        document.body.style.cursor = "none";
+                        isCursorVisible = false;
+                    }
+                };
+                
+                document.onmousemove = function() {
```
henrikingo: I think you should use document.addEventListener() instead. Other than that I agree with this approach.

js/impress.js
```diff
@@ -397,6 +397,33 @@
...
+            // Hide the mouse cursor in fullscreenmode after the configured
+            // amount of milliseconds without movement
+            if(rootData.hideMouseAfter) {
+                var mouseMoveTimer = null,
+                isCursorVisible = true;
+                
+                var hideCursor = function() {
+                    //isFullscreen
+                    if(!window.screenTop && !window.screenY){
+                        mouseMoveTimer = null;
```
henrikingo: I'm unsure about this line. If someone were to call hideCursor() some other way than via setTimeout() timeout triggering, then mouseMoveTimer will be null but the timeout will still expire? Wouldn't it be better to call window.clearTimeout(mouseMoveTimer) here?

Alternatively, I'm not sure you need to do anything about mouseMoveTimer here. Wouldn't this work just as well if you just leave this line?

js/impress.js
```diff
@@ -397,6 +397,33 @@
...
             
+            // Hide the mouse cursor in fullscreenmode after the configured
+            // amount of milliseconds without movement
+            if(rootData.hideMouseAfter) {
+                var mouseMoveTimer = null,
+                isCursorVisible = true;
+                
+                var hideCursor = function() {
+                    //isFullscreen
+                    if(!window.screenTop && !window.screenY){
```
henrikingo: This if() is actually a bit obscure. It would be more readable if you implement a helper function isFullscreen(), which returns true when this if is true. (I can get it since there is the comment, but would prefer the helper function, especially as this code moves to a separate plugin, so an extra line of code or two isn't an issue.)

js/impress.js
```diff
@@ -397,6 +397,33 @@
...
+                        isCursorVisible = false;
+                    }
+                };
+                
+                document.onmousemove = function() {
+                    if (mouseMoveTimer) {
+                        window.clearTimeout(mouseMoveTimer);
+                    }
+                    if (!isCursorVisible) {
+                        document.body.style.cursor = "default";
```
henrikingo: If the user wanted to theme the cursor with css, you will now unset that. You need to save the original cursor value when hiding it, then set it back to original value here.

henrikingo:
FYI, based on this idea, I've implemented a more generic `mouse-timeout` plugin in my repo. It supports hiding mouse cursor, but also hiding other things.

The [README](https://github.com/henrikingo/impress.js/blob/myfork/src/plugins/mouse-timeout/README.md) has the details, let me know what you think.

Compared to this PR:
 - The feature is still optional. It now depends on user writing CSS for the `body.impress-mouse-timeout` selector. It does not, however, add any `data-*` attribute.
 - I omitted the check for whether presentation is in full-screen mode, since IMO hiding the mouse cursor is ok to do also in a normal window. However, following the same style, it would be easy to add a separate `is-fullscreen` plugin, that would set a `body.impress-is-fullscreen` class when the window is in fullscreen mode.

sclausen:
@henrikingo your mouse-timeout-hide-stuff plugin seems to be superior and we should dump my PR in favour of yours.

FagnerMartinsBrack:
@sclausen We are still tracking this in the main impress repo.

henrikingo:
mouse-timeout plugin was merged to master

PR #546, opened 2016-04-01
s/bartaz/impress/g

perpi:
3 links are redirected from /bartaz/impress.js to /impress/impress.js
It made me a bit confused for a while! I was thinking if /impress/impress.js is the main account/repo, why you're keeping /bartaz/impress.js?!


bartaz:
Thanks, this should obviously be done at the time of moving the repository.

Issue #384, opened 2014-04-10
suppress auto advance when switching to browser window


adamralph:
Using chrome 34 on Windows 8.1, when I Alt+Tab away from the browser window and then Alt+Tab back to it, the presentation advances automatically to the next step.

Is there a way to suppress this behaviour?

Issue #304, opened 2013-06-30
ALT-TAB Problem


1mgtheboss:
I want to share this observation:
When I go to a separate window from Chrome which has an impress.js presentation
open and then come back to Chrome using ALT-TAB, it automatically goes to the
next slide.

I know that when showing a presentation naturally we don't go to other windows but,
there is no such restriction. So I think this is an unwanted side effect and should
be fixed.

Please share your opinions.

Issue #127, opened 2012-03-22
Alt-tab'ing into presentation in Chrome triggers step-next


blahbap:
When alt-tab'ing from another application into Chrome, a "step next" is triggered, (next slide is shown). This behavior does not occurs in Firefox. This is somewhat annoying when holding a presentations and you're switching between 2 or more applications.

PR #550, opened 2016-04-16
Use shift + Tab to go back to the previous step

FagnerMartinsBrack:
There is no test for this and I am not sure that we can have a proper test soon. My proposal is to land things like this now and worry about testing it later.

If we do this way, then we can at least land small stuff that are not that critical but have a lot of requests.

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -724,16 +724,24 @@
         //   Well, the [tab] key by default navigates around focusable elements, so clicking
         //   it very often caused scrolling to focused element and breaking impress.js
         //   positioning. I didn't want to just prevent this default action, so I used [tab]
-        //   as another way to moving to next step... And yes, I know that for the sake of
-        //   consistency I should add [shift+tab] as opposite action...
+        //   as another way to moving to next step...
         document.addEventListener( "keyup", function( event ) {
+
+            // Handles Tab and Shift + Tab
+            if ( event.keyCode === 9 ) {
+              if ( event.shiftKey ) {
+                api.prev();
+              } else {
+                api.next();
+              }
+              return;
+            }
 
             if ( event.shiftKey || event.altKey || event.ctrlKey || event.metaKey ) {
                 return;
             }
 
-            if ( event.keyCode === 9 ||
-               ( event.keyCode >= 32 && event.keyCode <= 34 ) ||
+            if ( ( event.keyCode >= 32 && event.keyCode <= 34 ) ||
                ( event.keyCode >= 37 && event.keyCode <= 40 ) ) {
                 switch ( event.keyCode ) {
                     case 33: // Page up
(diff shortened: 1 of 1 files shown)


js/impress.js
```diff
@@ -724,24 +724,31 @@
...
+            if ( event.keyCode === 9 ) {
+              if ( event.shiftKey ) {
+                api.prev();
+              } else {
+                api.next();
+              }
+              return;
+            }
+
             if ( event.shiftKey || event.altKey || event.ctrlKey || event.metaKey ) {
```
bartaz: Please note that before any key was handled we check for `shirt`, `alt`, `meta` etc modifiers and don't do anything if they are used.

This is a fix for an issue for example with windows Alt+Tab or mac Cmd+Tab for switching windows that was causing slides to move.

With your change bug will be back - Alt+Tab and Cmd+Tab will again cause slides to move when switching windows. Before you check for tab or shift tab you should make sure no other modifier keys are pressed.

FagnerMartinsBrack: I could find [the commit](https://github.com/FagnerMartinsBrack/impress.js/commit/d8995aa3194f6fdbcc8867626d52f913d0c52882), but not the original issue that describes the bug those lines fix.

Do you remember which browser and OS the problem happens? In Chrome on Mac, the event is not triggered when switching windows using <kbd>command</kbd> + <kbd>Tab</kbd>.

The only thing that is triggered is the following:

```shel
> event.keyCode: 91
> event.shiftKey: false
```

The code 91 represents OSLeft, which in Mac is the <kbd>command</kbd> key.

bartaz: It was reported couple of times: https://github.com/impress/impress.js/issues?utf8=%E2%9C%93&q=is%3Aissue+alt+tab

It seems it doesn't happen to everyone, so may be system and/or browser dependent.

FagnerMartinsBrack: Those are the ones who seems to be related to this:

* https://github.com/impress/impress.js/issues/127
* https://github.com/impress/impress.js/issues/238
* https://github.com/impress/impress.js/issues/304
* https://github.com/impress/impress.js/issues/384

I will take a look in those and try to reproduce it.

I wonder if it's worth keeping this issue in 0.6.0 milestone though, given we don't have a single test to back this up yet. It seems more complex than I thought and maybe it is not worth the risk of landing without tests for something that is even hard to reproduce.

FagnerMartinsBrack: Even when #510 is landed, we still won't be covering this.

FagnerMartinsBrack: Also, since we are just returning and not calling `event.preventDefault()`, this might break the focusable elements inside the step.

FagnerMartinsBrack:
I removed this from milestone 0.6.0, too risky.

js/impress.js
FagnerMartinsBrack: Here is probably an issue that was related to this #238.

henrikingo:
Shift+Tab now works in master.

Issue #167, opened 2012-06-08
On scroll/move event handler?


iluxav:
Hi,

Is there a event handler like onScroll or onMove? 

I'm trying to build a uml diagram presentation, where i have DIV's connected with lines between them. 
The problem is when I loading the page it draws lines perfect, even after scrolling to another div,  however i cannot find any event that could redraw the lines in real time(During the scrolling) and keep it connected even during the animation.

Thank you.

FagnerMartinsBrack:
Hi @iluxav , in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

To give a little bit of context, recently [a decision was made in the project](https://github.com/impress/impress.js/issues/435#issuecomment-177602861) to make the development more active and the first task is to clear up older issues like this one to see if [the OP](http://netforbeginners.about.com/od/internetlanguage/f/What-Is-OP.htm) is still interested in keep it going.

FagnerMartinsBrack:
Closing due to the lack of feedback from [the OP](http://netforbeginners.about.com/od/internetlanguage/f/What-Is-OP.htm). If you want to work in the PR again, feel free to comment here to open up the discussion again

PR #553, opened 2016-04-23
Added mouse wheel support.

mtancoigne:
Hi ! I added a mouse wheel support for Impress.js (as proposed in #167)

It's a bit sensitive, but you can have a demo here : http://codepen.io/mtancoigne/full/eZjGxm/

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -584,6 +584,16 @@
             return goto( next );
         };
 
+        // `current` API function returns the current step step
+        var current = function() { 
+            return steps.indexOf(activeStep);
+        };
+
+        // `total` API function returns the number of steps present
+        var total = function() {
+            return steps.length;
+        };
+
         // Adding some useful classes to step elements.
         //
         // All the steps that have not been shown yet are given `future` class.
@@ -657,7 +667,9 @@
             init: init,
             goto: goto,
             next: next,
-            prev: prev
+            prev: prev,
+            current: current,
+            total: total
         } );
 
     };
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
@mtancoigne Have you seen the following discussion: https://github.com/impress/impress.js/pull/51#issuecomment-3494748?

When I access you demo using Mac touchpad it has an awful UX, it jumps from the first slide to the third when it should jump from the first to the second. In the gif below you see what happens when I am using the touchpad using the left/right functionality (It might take a while to load, almost 6MB):

example impress

mtancoigne:
@FagnerMartinsBrack Crap... I only tested with my mouse. I'll take a look with a touchpad and try to do something better...

Thanks for your feedback...

mtancoigne:
Oh, and that don't work with firefox :/

I'll come back later with this one :)

FagnerMartinsBrack:
@mtancoigne Feel free to do it. But keep in mind that, since it is not trivial, we might only be able to land this once we have proper regression tests, at least for the algorithm.

Also, we are not even sure if this should be implemented in impress.js core itself or as a pluggable component. If it turns out to be really too hard to implement, we might as well wait until we have a proper API to make it possible attaching this easily in the consumer side.

(In the last case, solutions that show how this could be achieved without changing the core are welcome. We can document that somewhere.)

mtancoigne:
Ok, my mistake was to use a deprecated event, so now it works for firefox too. The only thing that bothers me is the number of mouse steps to use before changing of slide. 

3 was really too short, I tried with 20 wich was good on a touchpad, but on another computer this was too much, forcing me to make avbout two complete wheel scrolls...

Implementing this as a plugin should be good, as long as we can access the current slide from the API, and the numbers of slides.

FagnerMartinsBrack:
> Implementing this as a plugin should be good, as long as we can access the current slide from the API, and the numbers of slides.

Do you want to create an issue or link and existing one about that? It might be worth revisiting.

PR #555, opened 2016-04-27
ES6 and Unit Tests

mr-rampage:
My main goal to contribute to this project is to add test coverage. I started migrating to ES6 in order to make it easier to test the internal methods. As an early start, I've separate two utility methods into a ES6 utility module and wrote two test cases. 

I've also setup the build pipeline, but haven't included the linter because it currently breaks due to the ES6 code.

Hopefully, these are in line with the project. My goal is to remove one function from the `core.js` and add unit tests to it on a daily basis.


FagnerMartinsBrack:
Thank you for the help!

There is already a [Pull Request for including tests](https://github.com/impress/impress.js/pull/510) that will be landed probably this weekend. That Pull Request should be given priority because, besides testing the actual functionality instead of just unit tests, it already received a lot of review.

This PR was created from your master branch, we recommend creating a topic branch for each feature though instead of creating a big Pull Request for everything. That helps separating the concerns and creating atomic PRs.

Also, we need to create issues first to discuss any new code instead of start implementing it. The use of ES6 or not is something that needs to be discussed first for example. Even though we still haven't talked about ES6 in this project, we need to assess the tradeoffs of using a transpiler right now.

mr-rampage:
Awesome! Thanks for the insight. Will definitely keep an eye on this project and hopefully contribute in the future.

Issue #512, opened 2016-02-06
Current plan for impress.js: Set up the repository


FagnerMartinsBrack:
Here are some tasks that we will probably be doing in the incoming weeks to setup impress.js in a state where everything is focused well documented for future contributors and users of impress.js.

**Clean up**

The tasks here take in consideration [my comment in the state of the project](https://github.com/impress/impress.js/issues/435#issuecomment-177916049) and @bartaz [suggestions](https://github.com/impress/impress.js/issues/435#issuecomment-178468261) in the same thread.

- [x] Check all Pull Requests to find the ones where the OP is still interested in keeping track of it with the following message:

> Hi \<name of the poster>, in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

> To give a little bit of context, recently [a decision was made in the project](https://github.com/impress/impress.js/issues/435#issuecomment-177602861) to make the development more active and the first task is to clear up older PRs like this one to see if [the OP](https://medium.com/@fagnerbrack/what-does-op-means-on-github-8db958b604af) is still interested in keep it going.

If the OP does not want it anymore, just close it or create an issue for tracking purposes.

- [x] Check all issues and close duplicates. If one issue refers to more than one feature, create one issue for each. (X-reference everything with github linking so we can get the whole context from any endpoint.). If the issue is solved, close it.

- [x] Review all PRs and close the ones that doesn't make sense or the ones that are doing more than one thing after acknowledgement of the OP (or 1 week of no response).

- [x] Review all issues and close the ones that doesn't make sense after acknowledgement of the OP (or 1 week after no response).

At this point we will have only the essential issues/PRs open, all the others will be closed.

**Contributing**

- [x] Create a CONTRIBUTING.md file to guide new contributions (or filter existing ones). [Here is a nice example](https://github.com/js-cookie/js-cookie/blob/3f2b5e6884407c54b391483f39ddcd4c70f9243c/CONTRIBUTING.md). Add to the "WANT TO CONTRIBUTE?" section of the README.

- [x] Review the changelogs (Fill the changelog in the "Releases" page for every version released so far)

**Documentation**

- [x] Clear the wiki, removing projects that are no longer maintained and useless pages (Fixed on https://github.com/impress/impress.js/issues/524#issuecomment-188218591).
- [x] Document the currently supported API in the README (#563).
- [x] Document the support only for evergreen browsers that support CSS 3D transform (easier than always having to update the support) (#565)

**Release**

- [x] Create a new minor release
[...]

PR #563, opened 2016-05-21
API reference docs

FagnerMartinsBrack:
This starts the reference docs for impress.js as stated in the [plan](https://github.com/impress/impress.js/issues/512).

- [x] Document [`.past`, `.future`, `.present` classes](https://github.com/impress/impress.js/blob/a1d6475bd5f31b46586ae73a5d6e995d3ca4732b/index.html#L247-L257)
- [x] Add additional documentation for the [rotation axis](https://github.com/impress/impress.js/blob/a1d6475bd5f31b46586ae73a5d6e995d3ca4732b/index.html#L281-L286)
- [x] Document [`.impress-not-supported` API](https://github.com/impress/impress.js/blob/a1d6475bd5f31b46586ae73a5d6e995d3ca4732b/index.html#L116)
- [x] Document the body classes `.impress-on-step-x` and `impress-on-step-id`: https://github.com/impress/impress.js/issues/564#issuecomment-220782097

@bartaz, this is important because it represents the public contract for consumers of the framework. We should be safe to change anything that is not publicly documented. See [here](https://medium.com/@fagnerbrack/library-internals-do-not-exist-1ff50a30e3b0) for a detailed information of why it is very important to have a public API that doesn't rely solely in code comments.


FagnerMartinsBrack:
@impress/mergers Please review.

Pierstoval:
I don't know if this comment has its place here or in another issue but...

What's the difference between `.present` and `.active` exactly?

FagnerMartinsBrack:
> I don't know if this comment has its place here or in another issue but...
> What's the difference between .present and .active exactly?

That's actually a very good question!

Looking at the code, the `.present` class is added and removed when the events [`impress:stepenter`](https://github.com/impress/impress.js/blob/a1d6475bd5f31b46586ae73a5d6e995d3ca4732b/js/impress.js#L610) and [`impress:stepleave are executed`](https://github.com/impress/impress.js/blob/a1d6475bd5f31b46586ae73a5d6e995d3ca4732b/js/impress.js#L614).

The `impress:stepenter` is executed [whenever the step is selected](https://github.com/impress/impress.js/blob/a1d6475bd5f31b46586ae73a5d6e995d3ca4732b/js/impress.js#L565) via `goto`, either manually or programatically. The `impress:stepleave` is the same, except that it is executed when [leaving the step](https://github.com/impress/impress.js/blob/a1d6475bd5f31b46586ae73a5d6e995d3ca4732b/js/impress.js#L494).

The `.active` class seems to be added and removed [**before** any of those events are executed](https://github.com/impress/impress.js/blob/a1d6475bd5f31b46586ae73a5d6e995d3ca4732b/js/impress.js#L454).

So, from all I can tell, the `.active` class and any animation based on it will start executing **before** the `.present` class is added.

In the blame, I can see that the original commit that added the `.present` feature was this one: https://github.com/impress/impress.js/commit/b23506e885d33b53191dc0675c0a05585895a24a. The [version 0.5 changelog](https://github.com/impress/impress.js/releases/tag/0.5) doesn't say much.

The API was added in the version 0.3.0 in [this commit](https://github.com/impress/impress.js/commit/c46d39b0de9da0b66c7cbf18fcae08e363c417d7), which represents the [first version of impress.js](https://github.com/impress/impress.js/blob/0.1/js/impress.js#L210).

**Conclusion**

Pierstoval:
In the docs, I think there should be a small explanation of the list of all events/things that happen when slides change (when clicking on a step element, or using `.goto()` or `.next()`, etc.), to be more clear about what happens

FagnerMartinsBrack:
@Pierstoval Totally agree.

- [ ] Add documentation for the `impress:stepenter` and `impress:stepleave` events.

FagnerMartinsBrack:
- [ ] Add documentation for the `impress:init` event.

FagnerMartinsBrack:
@impress/mergers @Pierstoval Ready for the next round of review.

I think we should be conservative on what to document. If we document something, then we need to take care not to break it in a new release, and since the tests are too weak there is no ensurance that it will not break.

The best thing to do here is document only what is being used in the wild. The best thing to do is to look for some forks and see how they are using impress.js. If there is a considerable amount of usage and the code comments mention that, then we should document it in the public API.

Ping @bartaz

DOCUMENTATION.md
```diff
@@ -0,0 +1,305 @@
...
+      inspired by the idea behind <a href="http://prezi.com">prezi.com</a> <br>
+      and based on the <strong>power of CSS3 transforms and transitions</strong> in modern browsers.
+    </p>
+</div>
+```
+
+
+#### 3D Coordinates Positioning (data-z)
+
+Define the pixel based position in which the **center** of the [Step Element](#step-element) will be positioned inside the infinite canvas on the third dimension (Z) axis. For example, if we use `data-z="-3000"`, it means that the [Step Element](#step-element) will be positioned far away from the camera (by 3000px).
```
FagnerMartinsBrack: I am using the concept "positioned inside the infinite canvas" when we talk about the actual element that is positioned. When we talk about perception of positioning, I use "positioned from the camera". Not sure if we should consistently use one or another instead.

DOCUMENTATION.md
```diff
@@ -0,0 +1,305 @@
...
+```JavaScript
+var overview = document.getElementById( "overview" );
+var api = impress();
+api.init();
+api.goto( overview );
+```
+
+Accepts an optional [Number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number) in the last argument that represents the duration of the transition in milliseconds. If not provided, the default transition duration for the presentation will be used.
+
+Triggers the `impress:stepenter` event in the [Root Element](#root-element) when the presentation navigates to the target [Step Element](#step-element).
```
FagnerMartinsBrack: So, you can add this listener either in the body or in the root element. Nor sure where this should be but it seems to make sense documenting that it should be in the Root Element, because we don't want to unnecessarily add things to the `body`, which makes extension of impress.js pretty hard (you can have only one `body` per page, and that means only one instance per page).

DOCUMENTATION.md
```diff
@@ -0,0 +1,305 @@
...
+});
+```
+
+Triggers the `impress:stepleave` event in the [Root Element](#root-element) when the presentation navigates away from the current [Step Element](#step-element).
+
+**Example:**
+```JavaScript
+var rootElement = document.getElementById( "impress" );
+rootElement.addEventListener( "impress:stepleave", function() {
+  console.log( "Left the current Step Element" );
```
FagnerMartinsBrack: I didn't find a way to retrieve the current step from the `impress:stepleave` event.

DOCUMENTATION.md
```diff
@@ -0,0 +1,305 @@
+# Reference API
+
+## HTML
+
+### Root Element
+
+impress.js requires a Root Element with id "impress". All the content of the presentation will be created inside that element. It is not recommended to manipulate any of the styles, attributes or classes that are created by impress.js inside the Root Element after initialization.
```
bartaz: The id of root element doesn't have to be "impress" (but is "impress" by default). It is configurable in impress "constructor".
For example:

```html
<div id="presentation"> ... </div>
<script>
impress("presentation").init();
</script>
```

In theory there can be multiple root elements on a page with multiple instances of impress being initialized on them.
But at the moment keyboard events will work on all of them at the same time.

DOCUMENTATION.md
```diff
@@ -0,0 +1,305 @@
...
+
+**Example:**
+
+```CSS
+.impress-not-supported .step {
+  display: inline-block;
+}
+```
+
+## JavaScript
```
bartaz: While usage based on default `impress()` is most common, I'd add a section explaining this "constructor" (or maybe more like a factory function) of impress.js.

`impress( rootId )` is a function that builds API for presentation with given id of root element ('impress' by default).

To avoid calling it all the time it's worth storing it in a variable:
```
var api = impress("my-root");
```

It's possible to have multiple instances of impress.js (with separate roots) on the same page:

```
var firstAPI = impress("first-root");
var secondAPI = impress("second-root");

firstAPI.init();
secondAPI.init();
```

DOCUMENTATION.md
```diff
@@ -0,0 +1,305 @@
...
+
+**Example:**
+
+```JavaScript
+var api = impress();
+api.init();
+api.prev();
+```
+
+### impress().goto(stepIndex|stepElementId|stepElement, [duration])
```
bartaz: I would add spaces `goto( stepIndex | stepElementId | stepElement, [duration] )` to make it more readable.

DOCUMENTATION.md
bartaz: I guess right now we depend on step classes. Maybe it would be worth adding more data to events (for example both steps for example as `fromStep`, `toStep` or `currentStep`, `previousStep`).

bartaz:
@FagnerMartinsBrack 

Sorry for keeping you that long without any answer. I added couple of inline comments. But these are not big things.
I think these docs are already in a very good state and could be merged. We can work on the details later on. For sure it doesn't seem that there is anything wrong with the docs that would make something confusing or not working as expected.

DOCUMENTATION.md
FagnerMartinsBrack: I guess it might be a good idea for the future, but for now it might be better documenting only what actually exists without focusing in additional things. Is there any way, even with step classes, to be able to retrieve the current step from the `impress:stepleave` event?

DOCUMENTATION.md
FagnerMartinsBrack: > The id of root element doesn't have to be "impress" (but is "impress" by default). It is configurable in impress "constructor".
> For example:

I will edit that to consider those documented use cases.

> In theory there can be multiple root elements on a page with multiple instances of impress being initialized on them.
> But at the moment keyboard events will work on all of them at the same time.

Is multiple root elements on a page actually supported? If it is not explicitly commented in `index.html` shouldn't we just refrain from supporting it unless someone asks to do so?

People over time asked a lot for substeps, in this case if we implement substeps we might automatically fix the use cases for multiple instances of impress in the page and then we could manage to support it.

Does that make sense?

FagnerMartinsBrack:
@bartaz 

Sorry to take this longer to respond, I have been in a crazy marathon of events, meetups and trips this past few weeks.

I will handle those issues probably this weekend and merge it. In the meantime, a followup with some suggestion I have made there (to see if they make sense) would be much appreciated.

bartaz:
@FagnerMartinsBrack it's nothing compared to my response time, so no worries ;)

DOCUMENTATION.md
bartaz: I'm right now not 100% sure what would `.active` be during `stepleave` - in theory it should be always 'current' step.

But, step events are triggered on step elements. So it should be accessible always as `event.target`.

DOCUMENTATION.md
bartaz: @FagnerMartinsBrack I'm not sure how substeps connect to multiple impress instances.

Impress.js 'core' (the `impress()` factory function) supports multiple roots - meaning, you can create separate API objects for multiple root elements on the page, having multiple root elements initialised as impress presentations etc.

But of course global classes on body element may collide, styling and positioning may not work, etc. So even that it's technically possible I would probably not advertise having multiple roots on one page, because the result may be very unpredictable.

But for sure `impress("something")` can be used to initialise impress.js on element with different id than "impress" and it should be mentioned in docs.

DOCUMENTATION.md
FagnerMartinsBrack: Relying on `event.target` can be tricky because as far as I remember it represents the actual element on top of the mouse, not the one that contains the event handler, so the user needs to traverse back to find the desired element.

I will take a look one more time when I get into this PR again, maybe if `.active` doesn't work the `event.target` could be a temporary solution to be used in the example.

DOCUMENTATION.md
FagnerMartinsBrack: > I'm not sure how substeps connect to multiple impress instances.

I mean that it might fix the same use case that multiple impress instances in a page is trying to solve. Why one would want multiple impress.js instances in the page if they share the same keys as you said earlier?

> But of course global classes on body element may collide, styling and positioning may not work, etc. So even that it's technically possible I would probably not advertise having multiple roots on one page, because the result may be very unpredictable.

I will assume that is an argument for not documenting it then? :)

> But for sure impress("something") can be used to initialise impress.js on element with different id than "impress" and it should be mentioned in docs.

Sure, I will do that 👍

DOCUMENTATION.md
bartaz: Relying on `event.target` is tricky for click events (because browser triggers them for the top element touched by mouse). But impress.js events are programmatically called on step elements by impress.js, so as far as I remember `event.target` should always be step on which given event was triggered. So it's not a 'workaround', but a feature.

Only additional info (such as nextStep, etc) may be needed as additional data on event object.

FagnerMartinsBrack: That makes sense.

https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget:

> Identifies the current target for the event, as the event traverses the DOM. **It always refers to the element the event handler has been attached to** as opposed to event.target which identifies the element on which the event occurred.

I guess what we want here is `event.currentTarget` though.

bartaz: `currentTarget` is where handler is attached. So:

```js
rootElement.addEventListener( "impress:stepleave", function(event) {
    // event.target === step
    // event.currentTarget === rootElement
});
```

FagnerMartinsBrack: I manually tested your suggestion with `event.target === step` and it works! I will add that to the example 👍 .

DOCUMENTATION.md
FagnerMartinsBrack: > While usage based on default impress() is most common, I'd add a section explaining this "constructor" (or maybe more like a factory function) of impress.js.

👍

FagnerMartinsBrack:
> I think these docs are already in a very good state and could be merged. We can work on the details later on. For sure it doesn't seem that there is anything wrong with the docs that would make something confusing or not working as expected.

Landing on master then...

PR #565, opened 2016-05-29
Document support only for evergreen desktop browsers based on feature detection

FagnerMartinsBrack:
I don't think we should be focusing effort in maintaining this documentation. Stating that it works in the latest browsers and increase the support as the need arises based on developers feedback is the best thing we can do for now IMHO.

This also reduces a lot of content in the README, making it possible to focus in more important stuff than with needless details.

Ping @impress/mergers for review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,69 +64,11 @@ It's an (un)fortunate coincidence that a Open/LibreOffice presentation tool is c
 BROWSER SUPPORT
 -----------------
 
-### TL;DR;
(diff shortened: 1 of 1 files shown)


bartaz:
Yes, this sounds reasonable. This information in README is long outdated.

One thing that it would be worth adding/keeping is some information about mobile support, which is hard even though mobile browsers do support all needed APIs and are evergreen.
Of course, mobile support (for example should it be whitelisted) is a bigger topic for a separate discussion.

PR #566, opened 2016-06-06
Run the tests using Karma

FagnerMartinsBrack:
This improves the `npm test` command that was manually running the locally installed Chrome browser with the package [`chrome`](http://npmjs.com/package/chrome). That package also doesn't seems to be cross-platform, it looks like it [only works when using Mac](https://github.com/afc163/chrome/blob/e3720fc584c95eb79dea2e5b4c5cd9c6726176a1/index.js#L9).

Thanks @nkbt for helping me out to setup Karma in [CampJS VII](http://vii.campjs.com/)!


nkbt:
👌

PR #593, opened 2016-10-24
Move creds to circle-ci accnt config.

jdalton:
This PR moves creds to the circle-ci accnt config to simplify things and make it work for PRs.
Related to #91. Sauce creds are pretty low worry so the ways things are is a little overkill.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -2,10 +2,6 @@ machine:
   node:
     version: 6
 
-dependencies:
-  pre:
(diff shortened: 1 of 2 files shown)


FagnerMartinsBrack:
Do we know exactly the reason why the build is failing or we are just removing it to save us the trouble of having to look for it?

jdalton:
This should save us the trouble and avoids PR sniff logic.

Issue #476, opened 2015-05-19
IE perspective setting bug


bartuer:
In IE, transform : perspective(2000px) doesn't work, need set perspective on #impress directly, like
-ms-perspective : 2000px
![no_perspective](https://cloud.githubusercontent.com/assets/25855/7703762/b2313a9a-fe6c-11e4-9399-dff15518ed0e.PNG)
![has_perspective](https://cloud.githubusercontent.com/assets/25855/7703771/bc9403b4-fe6c-11e4-8256-1bde42bc104f.PNG)

please fix it

bartaz:
Thanks for reporting.

So you mean that `perspective` is not supported as a transform function in IE but works fine as separate value?

bartuer:
exactly.

jdalton:
Hey @bartuer!

Can you check out the PR over at #476 to see if this resolves the problem.

jdalton:
Erm over at #594.

jdalton:
Looks like it does 🎉

jdalton:
Moved to #594.

PR #594, opened 2016-10-26
Break-out perspective for better IE11 support.

jdalton:
This PR is an attempt to resolve the IE perspective issue described in #476.

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -133,11 +133,6 @@
         return " scale(" + s + ") ";
     };
 
-    // `perspective` builds a perspective transform string for given data.
-    var perspective = function( p ) {
-        return " perspective(" + p + "px) ";
-    };
-
     // `getElementFromHash` returns an element located by id from hash part of
     // window location.
     var getElementFromHash = function() {
@@ -384,7 +379,8 @@
             css( root, {
                 top: "50%",
                 left: "50%",
-                transform: perspective( config.perspective / windowScale ) + scale( windowScale )
+                perspective: ( config.perspective / windowScale ) + "px",
+                transform: scale( windowScale )
             } );
             css( canvas, rootStyles );
 
@@ -504,9 +500,13 @@
             // that both of them are finished.
             css( root, {
 
+                // For IE 11 support we must specify perspective independent
+                // of transform.
+                perspective: ( config.perspective / targetScale ) + "px",
+
                 // To keep the perspective look similar for different scales
                 // we need to 'scale' the perspective, too
-                transform: perspective( config.perspective / targetScale ) + scale( targetScale ),
+                transform: scale( targetScale ),
                 transitionDuration: duration + "ms",
                 transitionDelay: ( zoomin ? delay : 0 ) + "ms"
             } );
--- a/test/core_tests.js
+++ b/test/core_tests.js
@@ -37,9 +37,10 @@
     });
 
     QUnit.test( "Attributes", function( assert ) {
-      assert.expect( 9 );
+      assert.expect( 10 );
 
       var actual, expected;
+      var root = document.querySelector( "#impress" );
       var canvas = document.querySelector( "div#impress > div" );
 
       var canvasIsNotAStep = !canvas.classList.contains("step") && canvas.id === "";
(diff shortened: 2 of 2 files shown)


FagnerMartinsBrack:
If this fixes the problem, is there any way to build a failing test for IE to prevent regression?

jdalton:
<table>
<tr>
<th>before</th><th>after</th>
</tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

jdalton:
Based on the before and after it looks like it fixes the issue without regressing others (I checked Chrome at least) and without relying the `ms-` prefix.

I'll look into adding a test.

FagnerMartinsBrack:
Eventually, we need to get visual testing, but while we're not there if we can fail the test for IE by trying to retrieve the perspective data and then fix with this change then it would be a considerably decent regression test given the circumstances...

jdalton:
Yep, no worries. I'll cook up a test and a code comment prolly won't hurt either.

jdalton:
Tests added.

js/impress.js
```diff
@@ -133,11 +133,6 @@
         return " scale(" + s + ") ";
     };
 
-    // `perspective` builds a perspective transform string for given data.
-    var perspective = function( p ) {
-        return " perspective(" + p + "px) ";
-    };
-
```
FagnerMartinsBrack: I agree it's a good thing remove the function here since it doesn't have good cohesion (`perspective()` says nothing really). Sometimes no abstraction is better than the wrong abstraction 👍

test/core_tests.js
```diff
@@ -37,9 +37,10 @@
     });
 
     QUnit.test( "Attributes", function( assert ) {
-      assert.expect( 9 );
+      assert.expect( 10 );
 
       var actual, expected;
+      var root = document.querySelector( "div#impress" );
```
FagnerMartinsBrack: Is there a reason for using `div#impress` instead of just `impress` in the selector? I suspect this tried to keep consistency but the line below seems to have selected the `html` by mistake, is that right?

test/core_tests.js
```diff
@@ -78,6 +79,11 @@
       expected = "ease-in-out";
       assert.strictEqual( actual, expected, "canvas.style.transitionTimingFunction initialized correctly" );
 
+      // Ensure we specify perspective independent of transform for IE 11
```
FagnerMartinsBrack: Maybe if we change the message for something like `"root.style.perspective should be set explicitly for IE 11"` then we wouldn't need the comment and it would show the "why" when the test fails in the console.

The other tests are like "`n` initialized correctly" because we don't really had a reason for testing that attribute, it was just to have some blind coverage somewhere (which actually worked since when we added SauceLabs the tests failed in Safari xD).

test/core_tests.js
```diff
@@ -78,6 +79,11 @@
...
+      actual = root.style.perspective;
+      expected = "";
+      assert.notStrictEqual( actual, expected, "root.style.perspective initialized correctly" );
```
FagnerMartinsBrack: It looks like this test wouldn't fail only in IE but just serve as a verification that if we remove the `perspective` property it will fail in any browser, is that right?

jdalton: Correct.

js/impress.js
```diff
@@ -205,9 +200,7 @@
         height: 768,
         maxScale: 1,
         minScale: 0,
-
```
jdalton: Noticed a format inconsistency.

test/core_tests.js
jdalton: Ya I did it for consistency with other references.

js/impress.js
FagnerMartinsBrack: Can JSCS or JSHINT handle this? It might be more useful if we could automate this kind of stuff to reduce the chances it's spotted in unrelated PRs.

test/core_tests.js
```diff
@@ -49,7 +50,7 @@
       expected = "rotateZ(0deg) rotateY(0deg) rotateX(0deg) translate3d(1000px, 0px, 0px)";
       assert.strictEqual( actual, expected, "canvas.style.transform initialized correctly" );
 
-      // Normalize result for IE 11 and Safari.
+      // Normalize result for IE 11 and Safari
```
FagnerMartinsBrack: Can JSCS or JSHINT handle this? It might be more useful if we could automate this kind of stuff to reduce the chances it's spotted in unrelated PRs.

FagnerMartinsBrack: (I wonder if it's not more useful to do this kind of change in a separate PR, though?)

jdalton:
Updated.

PR #596, opened 2016-11-04
Proposal of documenting a high-level Roadmap

FagnerMartinsBrack:
I have created [a Roadmap in the wiki](https://github.com/impress/impress.js/wiki/Roadmap) based on what we have discussed so far since 2013 and I would like to propose documenting that so that the strategy is clear for future contributors.

Is the Roadmap correct? Should we document this?

Looking for feedback.

Ping @impress/mergers

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,3 +1,10 @@
+Roadmap
+-------
+
+impress.js have [an strategic Roadmap](https://github.com/impress/impress.js/wiki/Roadmap) with a high-level overview of what we plan for the future. When proposing changes for improvement, make sure it goes in that direction, even if it's a very small step.
+
+Since this is entirely a community effort, feel free to [create an issue](https://github.com/impress/impress.js/issues) and propose changes on that Roadmap.
+
 Issues
 ------
 


carywreams:
#### periodic ( annual ? ) pros/cons review for typescript re-write
(not meant to be an incendiary request)

FagnerMartinsBrack:
@carywreams Thanks for the suggestions.

This project don't have enough resources to promise periodic releases. Releases happen as they go when we believe there are enough changes to justify a bump (either patch, minor or major) according to the rules of [Semantic Versioning 2.0.0](http://semver.org/)

About typescript, there's no plan in adding transpiler to impress.js. In a few years, TypeScript will become obsolete too as everything else so we don't want to tie ourselves to any dependency that we can't unplug. @bartaz [has made it clear in the past](https://github.com/impress/impress.js/issues/435#issuecomment-178468261):

> I know it can be tempting to put a load of tools into the project (build, transpilers, and what not), but I would love to keep impress.js as simple as possible, the 'old school' way: "here is a JS file, add it to HTML, init impress.js and it works"; I know that at least testing may require some dev tooling, but let's keep it to bare minimum

FagnerMartinsBrack:
bump

henrikingo:
This patch no longer merges cleanly, closing. 

I also updated the wiki page to say that the roadmap was largely achieved!

PR #604, opened 2017-01-29
Traduction

ArtNous:
Add traduction in Spanish.


ArtNous:
all right!  I'll try make an issue in Spanish and see how it works.  thanks.

henrikingo:
It's not scalable to translate code comments to every language like this. I propose you'd simply translate impress.js to Spanish in your fork, and we could link to such translations from the README and/or wiki.

PR #616, opened 2017-05-07
changing link fonts.googleapi.com  from http to https

darkshell:
linking google fonts from http to https


bartaz:
Good catch, thanks!

(and sorry it took us so long) ;)

PR #621, opened 2017-09-16
Merge more testing coverage from henrikingo fork

henrikingo:
* core_tests.js and navigation_tests.js have more tests
* qunit_test_runner.html runs QUnit in a browser (without karma) and
  provides a generic iframe based platform to test different  presentations.
* copy qunit.js and syn.js into the repo. This allows qunit_test_runner.html
  to work without any dependency on node/npm. (Karma obviously does need them.)
* Move jscs cli option to .jscsrc instead


henrikingo:
A few comments for discussion, beyond what the commit message says:

Perhaps the debatable thing where I have deviated from upstream here is the embedding of qunit.js and syn.js into the repo itself. (6881 lines) At the time, I wanted to avoid introducing a dependency on node/npm. As impress.js is purely client side, and also we had this desire to keep the repo small and standalone without dependencies on other libraries. (Primarily runtime, but I considered that also when doing testing.) Also, at that time it still wasn't settled whether one should use npm, bower or god knows what for client side javascript projects, so it felt meaningful to just keep it simple and avoid the question. 

Once karma and linting is used, we of course need npm and node anyways, but I decided to keep this approach so that it is still possible to develop and run tests simply by opening qunit_test_runner.html in a browser.

I don't have a strong opinion either way, but the above was my subjective preference. It's reasonable to say that qunit.js and syn.js should instead be consumed via node_modules.

As mentioned in commit message, the testing setup is different too. Upstream uses just a simple inline HTML in test/bootstrap.js, but this isn't scalable for adding more tests. qunit_test_runner.html instead will load any html file into an iframe, and tests are written such that they run outside the iframe.

.jscsrc
```diff
@@ -0,0 +1,5 @@
+{
+	"preset": "jquery",
+        // Since we check quotemarks already in jshint, this can be turned off
+        "validateQuoteMarks": false
```
bartaz: Is this still needed?

As far as I understand JSCS was merged with eslint, so we should need only one config file.

If that's a quick fix maybe we could do it in this PR, if not consider if for some cleanup later.

package.json
```diff
@@ -22,18 +22,20 @@
...
   "scripts": {
-    "lint": "jshint js/impress.js *.js test/bootstrap.js && jscs js/impress.js *.js test/bootstrap.js --preset=jquery",
+    "lint": "jshint js/impress.js test/*.js && jscs js/impress.js test/*.js",
     "test": "karma start --single-run",
     "test:dev": "karma start",
     "test:sauce": "karma start karma.conf-sauce.js"
   },
   "devDependencies": {
     "chrome": "0.1.0",
+    "firefox": "*",
```
bartaz: Is this dependency needed? Do we use firefox from node, or does karma requires it?

.jscsrc
bartaz: I just realised we use jshint, not eslint - so never mind. Let's keep it as it is and consider eslint later.

henrikingo:
> Is moving qunit to npm achievable in this PR or does it add significant complexity?

Would have been trivial, if only I would learn to remember the use of "proxies:" section in karma.conf.js. But now I added a commit where both qunit.js and syn.js are removed from the repo, and only used as node modules. I will squash the commits when discussion is finished.

package.json
henrikingo: Upstream only use Chrome for testing with karma. Well, at least in karma.conf.js. I'm actually unfamiliar with the Saucelabs setup.

I wanted to test with Firefox too, so I've added it. This module is needed, and there's a line in karma.conf.js that also adds Firefox to the list of browsers.

Note that Chrome doesn't allow JavaScript to access the inside of an iframe when the protocol is file:///. So for opening qunit_test_runner.html in a browser (without karma), it only works with Firefox. So for this reason it's actually important to at least keep Firefox in the CI loop. Chrome is then kinda optional, but I like testing with both and I guess I also wanted to keep Chrome as upstream has been using that. 

And to make it crystal clear: both of them work fine over karma, because then files are served over http.

package.json
```diff
@@ -22,20 +22,23 @@
...
     "jshint": "2.9.1",
     "karma": "0.13.22",
     "karma-chrome-launcher": "1.0.1",
     "karma-cli": "1.0.0",
+    "karma-firefox-launcher": "~0.1",
     "karma-qunit": "1.0.0",
     "karma-sauce-launcher": "1.0.0",
-    "qunitjs": "2.0.0-rc1"
+    "qunitjs": "2.0.0-rc1",
+    "syn": "*"
```
bartaz: I'm a bit suspicious for using '*' as a version number (and simply allowing any even breaking update to be install).
I see that it's used on modules with major version 0 (which means that quite likely even minor changes may be breaking, so there is probably no need to lock on '0'), but for keeping updates safe maybe it would be better to lock the version to the exact version that it is now (and we know is working fine)?

test/core_tests.js
```diff
@@ -1,166 +1,266 @@
...
+    doneSync();
+    console.log( "End init() test (sync)" );
+  } ); // LoadIframe()
+
+} );
+
+// Note: Here we focus on testing the core functionality of moving between
+// steps, the css classes set and unset, and events triggered.
+// TODO: more complex animations and check position, transitions, delays, etc...
+QUnit.test( "Impress Core API", function( assert ) {
```
bartaz: Generally I would prefer the tests to be smaller (focused on one particular functionality or method) rather than one big test for whole 'core API', but I understand the complexity of the tests setup etc - and I suppose nature of impress.js doesn't make it very unit-test-friendly.

So I'm OK to leave it as it is for now not to block further merges, just wanted to leave it as a note for future consideration.

package.json
henrikingo: I can do that. For the record, they will be firefox:0.0.1 and syn:0.10.0

test/core_tests.js
henrikingo: Current breakdown is:
* init() tests
* next(), prev(), goto() = "movement" tests
* navigation module (testing key events and clicking a link with syn)
* navigation module no-op tests (e.g. clicking on current slide should not cause movement)

They are grouped into 2 files. They could be grouped into 4, but I don't know that that's what you mean.

The setup overhead is indeed quite high. So I don't see it meaningful to think that each "QUnit.test()" section would only test a single thing, I've rather used that as a kind of module, which will then do 30-100 different "actual tests" and related assertions.

PR #628, opened 2017-10-11
Add a framework for synchronously executed preInit and preStepLeave p…

henrikingo:


henrikingo:
Adds a pre-init and two pre-stepleave plugins. [The concept is already described in the README committed earlier](https://github.com/impress/impress.js/blob/dev/src/plugins/README.md#pre-init-plugins).

js/impress.js
```diff
@@ -659,6 +705,64 @@
         return lib;
     };
 
+    // `addPreInitPlugin` allows plugins to register a function that should
+    // be run (synchronously) at the beginning of init, before
+    // impress().init() itself executes.
+    impress.addPreInitPlugin = function( plugin, weight ) {
```
FagnerMartinsBrack: Why is weight necessary? Are there plugins who rely on the order of execution?

js/impress.js
```diff
@@ -659,6 +705,64 @@
...
+        if ( preInitPlugins[ weight ] === undefined ) {
+            preInitPlugins[ weight ] = [];
+        }
+        preInitPlugins[ weight ].push( plugin );
+    };
+
+    // Called at beginning of init, to execute all pre-init plugins.
+    var execPreInitPlugins = function( root ) { //jshint ignore:line
+        for ( var i = 0; i < preInitPlugins.length; i++ ) {
+            var thisLevel = preInitPlugins[ i ];
```
FagnerMartinsBrack: What's gonna happen when two pre-init plugins have the same `weight`?

js/impress.js
henrikingo: As a simple example, plugin-b might depend on the output (e.g. some DOM modifications) of plugin-a.

If you look at all the plugins in my repo, there's also other "logical" reasons to use some ordering. For example, substep plugin will prevent transitioning to next step, and instead reveals a substep on the same step. So it makes sense for that one to execute early, because it will in any case cancel the stepleave event.

With some other plugins it could be ambiguous/conflicting what should happen if user tries to use both, such as using both skip (skip next slide) and goto (jump to arbitrary slide). So the weight makes their priority explicit.

js/impress.js
henrikingo: The official answer is undefined. You should use different weights when 2 plugins have some inter-dependency, to ensure correct ordering.

They are pushed into an array and executed on a FIFO basis. So in practice the order of plugins in build.js determines the order when weight is equal.

FagnerMartinsBrack: What about scalability? Right now the building of plugins is centralized, so it's easy to keep track of the `weight` numbers. If we start exposing to the public it's unlikely that plugin authors will be able to keep track of the ordering.

Is there a way plugins can be created so that they don't have inter-dependencies?

js/impress.js
FagnerMartinsBrack: > For example, substep plugin will prevent transitioning to next step, and instead reveals a substep on the same step. So it makes sense for that one to execute early, because it will in any case cancel the stepleave event.

In the case of the substep plugin, instead of the plugin author to decide the priority of the execution, couldn't the user decide it? If you want to intercept the event you would put the plugin in a `script` tag further down the list and the plugin would do something like `event.stopPropagation()`:

```html
<script src="impress.js"></script>
<script src="goto-plugin.js"></script>
<script src="substep-plugin.js"></script> <!-- This has priority on the event related to navigation, because it will prevent the default behavior from the "goto" plugin -->
```

**impress.js**
```javascript
.addEventListener( "impress:stepleave", function( event ) {
  // Impress handlers
});
```

**goto-plugin.js**
```javascript
.addEventListener( "impress:stepleave", function( event ) {
  // goto-plugin handlers
});
```

**substep-plugin.js**
```javascript
.addEventListener( "impress:stepleave", function( event ) {
  event.stopPropagation();
  // stepleave handlers
});
```

js/impress.js
```diff
@@ -379,11 +386,23 @@
 
         // `goto` API function that moves to step given as `el` parameter (by index, id or element).
         // `duration` optionally given as second parameter, is the transition duration in css.
-        var goto = function( el, duration ) {
+        // `reason` is the string "next", "prev" or "goto" (default) and will be made available to
+        // preStepLeave plugins.
+        // `origEvent` may contain event that caused the calll to goto, such as a key press event
```
bartaz: Small typo in `calll`.

js/impress.js
```diff
@@ -379,11 +386,23 @@
...
+        var goto = function( el, duration, reason, origEvent ) {
```
bartaz: Right now it doesn't seem like `reason` or `origEvent` are needed - especially `reason` seems very arbitrary with 3 possible values.
But I assume they are necessary for some plugins to tell the difference of what triggered the transition?

js/impress.js
bartaz: I'm also not a big fun of this `weight` parameter, as it may be quite confusing for end users.
Luckily - this is quite 'internal' API used by plugin developers only now.

Yes, it would be great to have plugins being independent, but it's a theoretical ideal state that in practice would be very hard to avoid. Impress is a library working in browser DOM and modifying it, and plugins will rely and modify DOM, so there will be side-effects by definition and therefore plugins may conflict in some areas.

Moving freedom (and responsibility) of defining plugins order to impress.js user could be a 'solution' to this, but on the other hand it makes it harder for them to use plugins, because they would need to understand what plugins do internally. And really it's a plugin developer who knows best how it works and in what order it should be called. Yes, it has a downside of not scaling very well if external people would start to work on plugins, but honestly - we are really far from that and hopefully not many plugins would really need a control over when they need to be called (apart from internally developed plugins), so external developed plugins would be able to ignore weight altogether.

js/impress.js
```diff
@@ -659,6 +705,64 @@
...
+                for ( var j = 0; j < thisLevel.length; j++ ) {
+                    thisLevel[ j ]( root );
+                }
+            }
+        }
+    };
+
+    // `addPreStepLeavePlugin` allows plugins to register a function that should
+    // be run (synchronously) at the beginning of goto()
+    impress.addPreStepLeavePlugin = function( plugin, weight ) { //jshint ignore:line
```
bartaz: I don't like the fact that we are adding a lot of (kind of) internal api to global `impress` namespace, as it seems polluting it a bit...
I know that these methods need to be globally accessible, but maybe wrapping them in some `plugin` namespace would work? Like `impress.plugin.addPreStepLeavePlugin` etc?

As previously - not something I want to be changed now (knowing the dependency on existing plugins), but something to consider when we make new plugin API released and public.

src/plugins/goto/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+        <!-- When leaving this step, go directly to "step-5" -->
+        <div class="step" data-goto="step-5">
+
+        <!-- When leaving this step with next(), go directly to "step-5", instead of the next step.
+             If moving backwards to previous step - e.g. prev() instead of next() - then go to "step-1". -->
+        <div class="step" data-goto-next="step-5" data-goto-prev="step-1">
+
+        <!-- data-goto-key-list and data-goto-next-list allow you to build advanced non-linear navigation. -->
+        <div class="step" data-goto-key-list="ArrowUp ArrowDown ArrowRight ArrowLeft" data-goto-next-list="step-4 step-3 step-2 step-5">
```
bartaz: It's not clear that goto-key-list is a list of `KeyboardEvent.key` values. It would be good to mention what values can be used here for different keys - probably pointing to MDN documentation of `KeyboardEvent.key` would be enough.

js/impress.js
henrikingo: For example the goto plugin needs origEvent, because you can tie the pressed key to go to an arbitrary slide (e.g. press 5 to go to step-5). So it needs to have access to the event that generated the transition to know whether it was a keypress. stop plugin uses reason. It only stops next events, but doesn't prevent you going backwards to prev.

js/impress.js
henrikingo: Actually, I could make such change. The nr of plugins that use this api is small, in fact I think they are contained in this PR, so it is feasible to change.

I should note, in defense of this choice, that hey aren't added to the api returned by impress(). (Which would actually be incorrect anyway, since these are called before impress().init()). So they are kind of hidden aside already, and there are only two.


So you're suggesting:

impress.plugin.addPreInitPlugin()
impress.plugin.addPreStepLeavePlugin()

I actually think that's a bit confusing. These methods are offered by core impress.js. When I look at the above, it looks more like they *are* plugins, which is not the case.

But let me know if you insist, this is a change that can be done in this PR.

src/plugins/goto/README.md
henrikingo: Good point.

js/impress.js
henrikingo: The main point about weight is indeed that this is internal api to be used by plugins. Plugin developers will know how their plugin interacts with other plugins, if at all. There is no reason for presentation authors to be aware of it.

Because it is "internal", it can also be changed later if it's not sufficient. FWIW, Drupal plugins have a similar weight mechanism. So while it's simple, I believe it is scalable.

henrikingo:
I did the minor fixes in the new commit. Let me know what you want to do about addPreInitPlugin and addPreStepLeavePlugin.

js/impress.js
bartaz: I know that they are not part of api returned by `impress()` and it's good.
And also, now looking at this I agree that it may be interpreted as kind of a plugin, so I'm fine with leaving it as it is. Maybe later we can have a look on what all methods are on `impress` object and see if it makes sense to move or rename them somehow.

bartaz:
@henrikingo Thanks. I'm fine with what we have, feel free to merge.

js/impress.js
FagnerMartinsBrack: I go for @henrikingo on this one. The methods are built to create plugins so there's no reason to be anywhere else other than the core.

PR #631, opened 2017-10-21
Merge mobile stuff

henrikingo:
@bartaz : From now on this is basically just more plugins and more stuff. 

This first bucket contains a couple plugins that were hanging around in the PR queue. I have adapted them to the new plugin framework while still putting original author as author. Related to these plugins, the check that prevents from running on mobile phones is removed from core. (I'm aware you've been reluctant to do that, but this is a very common request from community. The mobile plugin attempts to help making presentations consume less memory when targeting mobiles.)

The swipe feature requires working with data structures in core. I've separated out that part, and it adds a new API function: impress().swipe(). The plugin part contains the UI facing part.

In the same PR some unrelated commits to core impress.js. After this, impress.js file is now identical to what I have in my fork.


henrikingo:
Sorry, thiis is against master. I'lll close and create a new one.

PR #632, opened 2017-10-21
Merge mobile stuff

henrikingo:
@bartaz : From now on this is basically just more plugins and more stuff. 

This first bucket contains a couple plugins that were hanging around in the PR queue. I have adapted them to the new plugin framework while still putting original author as author. Related to these plugins, the check that prevents from running on mobile phones is removed from core. (I'm aware you've been reluctant to do that, but this is a very common request from community. The mobile plugin attempts to help making presentations consume less memory when targeting mobiles.)

The swipe feature requires working with data structures in core. I've separated out that part, and it adds a new API function: impress().swipe(). The plugin part contains the UI facing part.

In the same PR some unrelated commits to core impress.js. After this, impress.js file is now identical to what I have in my fork.

### DIFF
--- a/src/impress.js
+++ b/src/impress.js
@@ -642,6 +642,10 @@
         // transition to happen, this is just to animate the swipe. Once the
         // transition is committed - such as at a touchend event - caller is
         // responsible for also calling prev()/next() as appropriate.
+        //
+        // Note: For now, this function is made available to be used by the swipe plugin (which
+        // is the UI counterpart to this). It is a semi-internal API and intentionally not
+        // documented in DOCUMENTATION.md.
         var swipe = function( pct ) {
             if ( Math.abs( pct ) > 1 ) {
                 return;
--- a/src/impress.js
+++ b/src/impress.js
@@ -54,6 +54,28 @@
 
     } )();
 
+    var validateOrder = function( order, fallback ) {
+        var validChars = "xyz";
+        var returnStr = "";
+        if ( typeof order === "string" ) {
+            for ( var i in order.split( "" ) ) {
+                if ( validChars.indexOf( order[ i ] >= 0 ) ) {
+                    returnStr += order[ i ];
+
+                    // Each of x,y,z can be used only once.
+                    validChars = validChars.split( order[ i ] ).join( "" );
+                }
+            }
+        }
+        if ( returnStr ) {
+            return returnStr;
+        } else if ( fallback !== undefined ) {
+            return fallback;
+        } else {
+            return "xyz";
+        }
+    };
+
     // `css` function applies the styles given in `props` object to the element
     // given as `el`. It runs all property names through `pfx` function to make
     // sure proper prefixed version of the property is used.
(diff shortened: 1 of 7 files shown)


src/impress.js
```diff
@@ -583,6 +615,110 @@
...
+        // Note that pre-stepleave plugins such as goto can mess with what is a
+        // next() and prev() step, so we need to trigger the pre-stepleave event
+        // here, even if a swipe doesn't guarantee that the transition will
+        // actually happen.
+        //
+        // Calling swipe(), with any value of pct, won't in itself cause a
+        // transition to happen, this is just to animate the swipe. Once the
+        // transition is committed - such as at a touchend event - caller is
+        // responsible for also calling prev()/next() as appropriate.
+        var swipe = function( pct ) {
```
bartaz: I think the name `swipe` may be a bit misleading. It took me a while to realise what this method is doing and why it needs to be in impress.js core (not in some external mobile plugin).

The name this function adds some 'swipe' (or touch) navigation. But it seems (as far as I understand) moves the transition between two steps (for example while user is swiping).

I feel a bit unsure about having it in main impress API, as it doesn't seem like it would be used by users, but rather then by plugins, but I don't really know how to approach it other way.

It it used by multiple plugins? If it's only for swipe maybe it would make sense to move it into mobile/swipe plugin that adds support for that?
I guess what would be needed for that in public API would be some  additional `move` method , that would allow impress.js to move into any given position (like `impress().move({x: 100, y: 200, z: 300, rotateX, scale...}`). But... that seems like other work I guess.

@henrikingo You probably have a bigger picture. Does it make sense to keep it in public API?

bartaz:
@henrikingo I'm ok with getting mobile check out of impress core. Mobile devices and browsers are way better then they were when impress.js was first created.
Still, on many it won't fully work, but lets give users an option to support mobile as they want.

src/impress.js
henrikingo: It's only used by the swipe plugin, so renaming or changing arguments is feasible here.

You're right that at this point your idea of a impress().move() function would really be handy, but it seems to me like a big change! In any case, as you can see from the first half of the swipe() function, it's not just about animating a swipe, but it's also related to the step concept. What really happens is a prestepleave -> swipe -> stepleave (or not, if cancelled). So what is now the swipe plugin, will still need to interact with a higher level part of the code that is in charge of the (pre)stepleave events, and only that part can then animate the actual swipe. From that point of view swipe() truly is an equal sibling to prev() and next().

henrikingo:
> lets give users an option to support mobile as they want

My thought exactly.

src/impress.js
bartaz: As it seems that the matter is more complicated, I don't mind leaving it as it is.
But I would try to think of some way to keep such methods more 'private' - at least not being so exposed to users.
I guess the easiest way would be to not document them in official public API, only as part of API docs (if that makes sense)? Or - if there will be more methods like that keep them separate from rest of public API?

henrikingo:
Cool. I added a code comment to make the point that swipe() is intentionally not documented in DOCUMENTATION.md for now.

PR #633, opened 2017-10-23
Merge all other plugins

henrikingo:
This adds all the other plugins into src/plugins, each in their own commits.

ping @bartaz

### DIFF
--- a/src/plugins/extras/README.md
+++ b/src/plugins/extras/README.md
@@ -1,0 +1,26 @@
+Extras Plugin
+=============
+
+The Extras plugin will initialize the optional addon plugins from 
+[extras/](../../../extras/) directory, if they were loaded.
+
+Generally, for an extras plugin to have been loaded, 2 things must have happened:
+
+1. The extras plugins must be present in extras/ directory, for example after 
+   running `git submodule update`
+2. One or more extras plugins are added to the impress.js presentation (the HTML
+   file) by the author using a regular `<script>` tag.
+
+If one or more extras plugins were so added, this plugin will automatically
+discover them and perform initialization (such as calling 
+`mermaid.initialize()`).
+
+If no extras plugins are added to a presentation, this plugin does nothing.
+
+Note that some extra plugins (like mathjax) initialize themselves immediately, and
+there's nothing to do here.
+
+Author
+------
+
+Henrik Ingo (@henrikingo), 2016
--- a/src/plugins/impressConsole/impressConsole.js
+++ b/src/plugins/impressConsole/impressConsole.js
@@ -1,0 +1,747 @@
+/**
+ * Adds a presenter console to impress.js
+ *
+ * MIT Licensed, see license.txt.
+ *
+ * Copyright 2012, 2013, 2015 impress-console contributors (see README.txt)
+ *
+ * version: 1.3-dev
+ *
+ */
+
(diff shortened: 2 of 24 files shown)


src/plugins/impressConsole/impressConsole.js
```diff
@@ -0,0 +1,747 @@
...
+        // Return the object
+        allConsoles[ rootId ] = { init: init, open: open, clockTick: clockTick,
+                               registerKeyEvent: registerKeyEvent };
+        return allConsoles[ rootId ];
+
+    };
+
+    // Returns a string to be used inline as a css <style> element in the console window.
+    // Apologies for length, but hiding it here at the end to keep it away from rest of the code.
+    var cssStyleStr = function() {
```
henrikingo: The original impressConsole has external CSS files. But we don't want, or even can't, have that for impress.js plugins. So I've simply inlined the default CSS here and it is added as a <style> tag above. Not the greatest thing I ever thought of, but this is a complete window with a complex GUI, there's no way to avoid the fact it needs its own CSS.

Note the use of `backtick quoted string` below, to create a multiline string. This is an ES6 feature.

src/plugins/impressConsole/impressConsole.js
```diff
@@ -0,0 +1,747 @@
...
+    // All console windows, so that you can call impressConsole() repeatedly.
+    var allConsoles = {};
+
+    // Zero padding helper function:
+    var zeroPad = function( i ) {
+        return ( i < 10 ? '0' : '' ) + i;
+    };
+
+    // The console object
+    var impressConsole = window.impressConsole = function( rootId ) {
```
henrikingo: Adds a global impressConsole object. This is against the rules for plugins, but allowed for backward compatibility. It is however considered deprecated, and a warning is logged if you try to call impressConsole().init().

src/plugins/skip/skip.js
```diff
@@ -0,0 +1,84 @@
...
+            event.detail.transitionDuration = util.toNumber(
+                event.detail.next.dataset.transitionDuration, event.detail.transitionDuration
+            );
+        }
+    };
+
+    // Register the plugin to be called in pre-stepleave phase
+    // The weight makes this plugin run early. This is a good thing, because this plugin calls
+    // itself recursively.
+    window.impress.addPreStepLeavePlugin( skip, 1 );
```
henrikingo: Somehow this was left out from the PR with the other preInit and preStepLeave plugins. It's here now.

src/plugins/extras/README.md
```diff
@@ -0,0 +1,26 @@
+Extras Plugin
```
henrikingo: Oh right: The extra addons themselves aren't yet in this branch. If you want to see some of them in action, see http://henrikingo.github.io/impress.js/examples/classic-slides/#/step-1

The point of this plugin is to simplify and standardize their use, so that if, for example, markdown.js was added with a script tag, then this plugin detects it and does whatever is necessary to convert markdown to html. The uses doesn't need to add additional `init()` or other lines of code to make it happen. In this sense, even if extra addons aren't part of impress.js, this still defines a small set of extra libraries who's use is supported and standardized.

henrikingo:
Based on earlier encouragement, merging this and moving forward to next PR (coming soon)

PR #637, opened 2017-10-30
Increased test portability using Chrome-Headless and updated dependen…

lguzzon:
…cies


karma.conf.js
```diff
@@ -1,3 +1,8 @@
+
+/*jslint node: true */
+
+process.env.CHROME_BIN = require( "puppeteer" ).executablePath();
```
henrikingo: Can you add a comment that explains why this is needed?

henrikingo:
Hi @Iguzzon

Thank you for your patience while I've been busy merging my own work upstream, and thank you also for already rebasing onto the new state of master. I like the idea of running karma with the browser in headless mode, especially since the qunit messages aren't visible in the browser window anyway.

Now that I added Firefox into the mix, it's kind of a shame to have Chrome in headless mode while Firefox would still popup. It looks like there's not an NPM plugin to run firefox headless in karma, but it would be simple to add a few lines to get there anyway: https://github.com/karma-runner/karma-firefox-launcher/issues/76

Are you interested in trying if that works?

Thanks also for upgrading the other modules to current versions.

karma.conf.js
lguzzon: Ok:
- Line 2 tells jslint that karma.conf.js is a "node.js" file so "linting" this file goes right 
- Line 4 set the right path to the chrome binary (puppeteer module bundle the chrome binary)

henrikingo: I was specifically talking about line 4. What I meant was, in your next patch, can you add a code code comment to the file. It should explain why we need to set CHROME_BIN, when previously it wasn't needed?

henrikingo:
Thanks for the update. To clarify, do you also want to try and make Firefox headless by following the example here: https://github.com/karma-runner/karma-firefox-launcher/issues/76

If not, I can merge what you have here and file a followup ticket for the firefox part. Just let me know.

henrikingo:
FYI @maxwell-k: looks like the conflict resolution breaks this PR and it didn't merge after all. The only thing that changed in the merge are some versions in package.json. It had previously passed a CircleCI build.

henrikingo:
This seems to be achieved by #722 . Thanks for your efforts though, this was a good idea and is about to be merged in other PR.

Issue #646, opened 2017-12-04
Link broken in plugins README


bcarter:
The link "extra addons" in the README at https://github.com/impress/impress.js/tree/master/src/plugins is broken (404).  I believe it should point to https://github.com/impress/impress-extras.  I'll submit a pull request to fix it.

PR #647, opened 2017-12-04
Fixed the link to "extra addons" in the plugins README

bcarter:
Fix for issue [#646](https://github.com/impress/impress.js/issues/646) changed the link in the README to point to https://github.com/impress/impress-extras

### DIFF
--- a/src/plugins/README.md
+++ b/src/plugins/README.md
@@ -21,10 +21,11 @@ Extra addons
 ------------
 
 Yet more features are available in presentations that enable 
-[extra addons](https://github.com/impress/impress-extras). Extra addons are 3rd party plugins integrated
-into impress.js to provide convenient and standardized access to them. However,
-they are not activated by default, rather must be included with a `<script>`
-tag.
+[extra addons](https://github.com/impress/impress-extras). Extra addons are 3rd party plugins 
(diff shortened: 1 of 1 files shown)


src/plugins/README.md
```diff
@@ -21,7 +21,7 @@ Extra addons
 ------------
 
 Yet more features are available in presentations that enable 
-[extra addons](../../extras/). Extra addons are 3rd party plugins integrated
+[extra addons](https://github.com/impress/impress-extras). Extra addons are 3rd party plugins integrated
```
henrikingo: Thanks for catching this!

In fact, this whole sentence is a bit out of date, it was written at a time when the extra addons were part of the main impress.js repo. While we're doing this, could you please also rewrite the immediately following sentence to something like this:

    Extra addons are 3rd party plugins that are not part of impress.js, but that we have nevertheless collected together into the [impress-extras](https://github.com/impress/impress-extras) repo to provide convenient and standardized access to them. To include the extra addons when checking out impress.js, use `git clone --recursive`. Even then, they are not activated by default in a presentation, rather each must be included with their own `<script>`
tag.

bcarter: Added the new sentence

henrikingo:
Thanks!

PR #672, opened 2018-02-20
PR for media plugin

complanar:
I’ve created a media plugin that enables pause on `impress:stepleave`, play on `impress:stepenter`, adds a special class to the body tag while playing media (useful for darkening the background in the meanwhile with CSS) and allows the use of some shorthands for inserting media. See commented source in `src/plugins/media/media.js` for details.

`js/impress.js` was not built so far.

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -2705,6 +2705,145 @@
 } )( document, window );
 
 /**
+ * Media Plugin
+ *
+ * The media plugin is a combination of a preeinit and prestepleave plugin. 
+ * It is executed before impress:init and before impress:stepleave.
+ * 
+ * It will do the following things:
+ * 
+ *   - Parse special shortcuts to full html5 audio or video nodes:
+ *     e.g. <div class="media" data-media-source="…" data-media-type="audio/…"></div>
+ *     or   <span class="media" data-media-source="…" data-media-type="video/…"></span>
+ *     
+ *     All type of html tags are allowed. Currently only a single source is supported.
+ *     All nodes will have the control attribute added, except the 'data-media-controls="false"'
+ *     attribute is added.
+ *     
+ *         <div class="media" data-media-source="…" data-media-type="…/…" data-media-controls="false"></div>
+ * 
+ *  - Add a special class when playing and pausing media (removing them when ending)
+ *
+ *  - Pause all video and audio an the active step when leaving it. This can be disabled for 
+ *    indivudual media nodes by adding the data-media-pauseonleave="false" attribute. This attribute
+ *    is inherited from the shortcut described above.
+ *    
+ *    Examples: 
+ *
+ *      <audio contols data-media-pauseonleave="false">
+ *        <source src="…" type="audio/…">
+ *        Your browser does not support HTML5 audio. Please update you browser to 
+ *        a recent version.
+ *      </audio>
+ *      
+ *      <div class="media" 
+ *           data-media-source="…" 
+ *           data-media-type="video/…" 
+ *           data-media-controls="false" 
+ *           data-media-pauseonleave="false">
+ *      </div>
+ *
+ * Copyright 2018 Holger Teichert (@complanar)
+ * Released under the MIT license.
+ */
+/* global window, document, impress */
+
+(function (document, window) {
+  "use strict";
+  
+  // function names
+  var parseMediaShortcut, 
+      enhanceMediaNode, 
+      enhanceMedia, 
+      pauseOnStepleave, 
+      onPlay, 
+      onPause, 
+      onEnded;
+
+  // Parse nodes with media class <div class="media" data-media-source="PATH" data-media-type="MIMETYPE"></div>
+  parseMediaShortcut = function (target) {
+    var data, mimeParts, controls, pauseonleave;
+    data = target.dataset;
+    if (data.mediaType && data.mediaSource) {
+      mimeParts = data.mediaType.split('/');
+      target.classList.add('media-' + mimeParts[0]);
+      controls = data.mediaControls === "false" ? '' : ' controls';
+      pauseonleave = data.mediaPauseonleave === "false" ? ' data-media-pauseonleave="false"' : '';
+      target.innerHTML = '<' + mimeParts[0] + controls + pauseonleave + '><source src="' + data.mediaSource + '" type="' + data.mediaType + '"> Your Browser does not support HTML5 ' + mimeParts[0] + '.</' + mimeParts[0] + '>';
+    }
+  };
+  
+  onPlay = function (event) {
+    var type = event.target.nodeName.toLowerCase();
+    document.body.classList.add(type + "playing");
+    document.body.classList.remove(type + "paused");
+  };
+  
+  onPause = function (event) {
+    var type = event.target.nodeName.toLowerCase();
+    document.body.classList.add(type + "paused");
+    document.body.classList.remove(type + "playing");
+  };
+  
+  onEnded = function (event) {
+    var type = event.target.nodeName.toLowerCase();
+    document.body.classList.remove(type + "playing");
+    document.body.classList.remove(type + "paused");
+  };
+  
+  enhanceMediaNode = function (type) {
+    var i, id, media = document.getElementsByTagName(type);
+    for (i = 0; i < media.length; i++) {
+      // Set an id to identify each media node - used e.g. by the consoleMedia plugin
+      id = media[i].getAttribute('id');
+      if (id === undefined || id === null) {
+        media[i].setAttribute('id', 'media-' + type + '-' + i);
+      }
+      media[i].addEventListener("play", onPlay, false);
+      media[i].addEventListener("playing", onPlay, false);
+      media[i].addEventListener("pause", onPause, false);
+      media[i].addEventListener("ended", onEnded, false);
+    }
+  };
+  
+  enhanceMedia = function () {
+    var i, media = document.getElementsByClassName('media');
+    for (i = 0; i < media.length; i++) {
+      parseMediaShortcut(media[i]);
+    }
+    enhanceMediaNode('audio');
+    enhanceMediaNode('video');
+  };
+
+  pauseOnStepleave = function (event) {
+    var videos, audios, media, type, i;
+    if ((!event) || (!event.target)) {
+      return;
+    }
+
+    videos = event.target.getElementsByTagName('video');
+    audios = event.target.getElementsByTagName('audio');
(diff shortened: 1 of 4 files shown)


src/plugins/media/media.js
```diff
@@ -0,0 +1,172 @@
...
+    videos = event.target.getElementsByTagName('video');
+    media = {
+      video: videos,
+      audio: audios
+    };
+    for (type in media) {
+      for (i = 0; i < media[type].length; i++) {
+        play = media[type][i].dataset.mediaPlayonenter;
+        console.log(play);
+        if (play !== undefined) {
```
henrikingo: && play !== "false"

src/plugins/media/media.js
```diff
@@ -0,0 +1,172 @@
...
+      media[i].addEventListener("ended", onEnded, false);
+    }
+  };
+  
+  enhanceMedia = function () {
+    var i, steps, media = document.getElementsByClassName('media');
+    for (i = 0; i < media.length; i++) {
+      parseMediaShortcut(media[i]);
+    }
+    enhanceMediaNode('audio');
```
henrikingo: You don't need to handle `audio` and `video` separately. You can get a list of both of them with

    medias = document.querySelectorAll('audio, video');

src/plugins/media/media.js
```diff
@@ -0,0 +1,172 @@
...
+        }
+      }
+    }
+  };
+
+  // Register the plugin to be called in the pre-init phase
+  window.impress.addPreInitPlugin(enhanceMedia);
+
+  // Register the plugin to be called in pre-stepleave phase
+  impress.addPreStepLeavePlugin(pauseOnStepleave);
```
henrikingo: The `pauseOnStepleave` (or equivalent, whatever we end up calling it) should be called on `impress:stepleave`. Just like `playOnStepenter` is triggered on `impress:stepenter`. It has no need to be a pre-stepleave plugin.

src/plugins/media/media.js
```diff
@@ -0,0 +1,172 @@
...
+    var i, steps, media = document.getElementsByClassName('media');
+    for (i = 0; i < media.length; i++) {
+      parseMediaShortcut(media[i]);
+    }
+    enhanceMediaNode('audio');
+    enhanceMediaNode('video');
+    steps = document.getElementsByClassName('step');
+    for (i = 0; i < steps.length; i++) {
+      console.log("Add event listener.");
+      steps[i].addEventListener('impress:stepenter', playOnStepenter, false);
```
henrikingo: Ok, here's the hard part...

impress.js now supports a `impress().tear()`. It will reset everything to its original state, as things were before `impress().init()` was called. This means that any changes your plugin does, must be recorded so that they can be reverted when `tear()` is called. This isn't documented in the plugin README, sorry about that.

To add event listeners, use `impress().lib.gc.addEventListener()`. Example: https://github.com/impress/impress.js/blob/master/src/plugins/form/form.js#L33

While on this topic, you must also create and register a callback function, that will remove any classes you added to `body`. Example: https://github.com/impress/impress.js/blob/master/src/plugins/rel/rel.js#L146

(In your case, callback function is simpler, just do a few `classList.remove()` calls and you're done.)

js/impress.js
```diff
@@ -2323,6 +2323,8 @@
 
             if ( consoleWindow && !consoleWindow.closed ) {
                 consoleWindow.focus();
+                triggerEvent(root, 'impress:console:open', {});
```
henrikingo: Please remember to run `node build.js` before submitting a PR. As far as I can tell, this `js/impress.js` now contains stuff that isn't at all part of this PR. This is more confusing than simply forgetting to re-build at all.

src/plugins/media/media.js
complanar: What exactly is the difference between a pre-stepleave plugin and binding it to `impress:stepleave`? I was already wondering, why there is no pre-stepenter plugin.

src/plugins/media/media.js
```diff
@@ -0,0 +1,172 @@
...
+ *
+ *  - Autostart videos when entering a step, if the attribute data-media-playonenter is set and not "false".
+ *
+ *  - Pause all video and audio an the active step when leaving it. This can be disabled for 
+ *    indivudual media nodes by adding the data-media-pauseonleave="false" attribute. This attribute
+ *    is inherited from the shortcut described above.
+ *    
+ *    Examples: 
+ *
+ *      <audio contols data-media-pauseonleave="false">
```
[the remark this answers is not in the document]
complanar: I very seldom use automatically starting media, so I am not the right person to decide which way is the most convenient for the user.

My most often use case is that:
- Enter a step
- Talk about something on the step
- Start media manually
- Sometimes talk a little bit more
- Leave the step, pausing or stopping all media

I think, it is very seldom that someone leaves a step but wants media to play along. Thus it should not be the default value to let it play, when it was started manually.

> If user re-enters the step, it should start media from the beginning. In other words, the stepleave action is to stop the media, not pause.

Are you sure? Maybe someone (including me) likes to proceed at the very same point with the playback. This is made impossible by stopping media instead of pausing it. Aside from that there is no official `.stop()` method for html5 media as far as I know, we would have to pause it anyhow and set the current position to 0.

I would vote for:
- data-media-autoplay: default false
- data-media-autopause: default true
- data-media-autostop: default true if autopause is true and autoplay is true otherwise false

We would have to define which of the two attributes autopause and autostop takes precedence over the other if opposing values were set. 
`data-media-autoplay="true" data-media-autopause="true" data-media-autostop="false"` would result in pausing media on stepleave and resuming on re-enter, according to my description above.
But what do we do on `data-media-autoplay="false"  data-media-autopause="true" data-media-autostop="true"`? Do later values overwrite earlier defined ones?

Of course one could also think of a different solution:
- data-media-autoplay: default false
- data-media-autostop: default copy from autoplay
- data-media-autopause: default true if autoplay is false and autostop is false otherwise false

Regardless of how it is done, it should be heavily documented, because it is complicated to understand if we make a difference between stopping and pausing, and which values must be set to get the desired behavious.

src/plugins/media/media.js
complanar: What about the defined IDs? (see `enhanceMediaNode()`) Should the be removed as well? How do I record which IDs were added and distinguish them from already existent IDs? Of course one could use a naming convention, but what if some user sets a id that matches that convention accidently? It would be removed as well.

src/plugins/media/media.js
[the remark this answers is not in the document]
complanar: Just realized that there would be no way to stop/pause a manually started media without returning to the containing step. I think this is highly inconvenient for users so should never be the default but require an explicit setting by the user.

src/plugins/media/media.js
henrikingo: The original idea for impress.js events is based on using javascript events. This is the preferred mechanism whenever suitable.

The pre- style events were added later to facilitate two use cases:
* Cases where the plugin really needs to run before the event it is tied to. For example, the rel plugin clearly must run before init itself.
* Cases where you may want to prevent the event from happening at all. For example, the stop plugin returns false, which stops the stepleave activity completely. (You will stay on the current step and not move to next.)

src/plugins/media/media.js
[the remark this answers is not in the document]
henrikingo: Ok, good suggestion to have both `autostop` and `autopause`.

One answer is easy for me to give: There is a general rule for plugins, that by default they should not do anything. So in this case, if user adds a video or audio tag to a step, and don't use any `data-` attributes, then user must manually start and stop the media. This is not negotiable. Think about it this way: user may not know of the existence of the media plugin, and we must not do anything surprising when it is not explicitly invoked via `data-media-` attributes.

Beyond that, I think I agree with the most common use case: If `data-media-autoplay="true"` this should imply that `data-media-autostop="true"` too. This way the media stops when leaving the step and starts from the beginning if you ever come back. (For example, a presentation might rotate around, in which case it is likely that you return to same slide again.)

*What should be the relation between autostop and autopause?*

IMO if both autostop and autopause are true, the logical result of this is "do both". In practice this is the same as autostop alone. You pause the media, then pause the media and return to beginning. (Or, pause and return to beginning, then pause.)

Again, doing this is my preference simply because it is what you would logically expect. It's an non-default edge case anyway, and you can get the behavior you want by explicitly setting true and false to get what you want.

***

So, as an example of how to achieve your preferred use case, you would do:

    <audio data-media-autoplay="false" data-media-autostop="false" data-media-autopause="true">

or simply:

    <audio data-media-autopause="true">

> Regardless of how it is done, it should be heavily documented, because it is complicated to understand if we make a difference between stopping and pausing, and which values must be set to get the desired behaviour.

Currently plugins are documented with a README.md file in the same directory with the rest of the plugin.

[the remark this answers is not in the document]
henrikingo: Btw, you don't need to implement any code for this, but in documentation you could add an example for how to automatically move to next step when the media reached its end:

    <audio onended="impress().next()">

This is a nice way of creating narrated presentations where presentation moves forward after the media clip finished playing. The above allows for tighter synchronization than for example setting step durations.

src/plugins/media/media.js
henrikingo: Yes. Anything you create should also be removed.

Unfortunately there isn't an easy `gc` helper for removing attributes. You need to create your own function and add it to the gc stack with `gc.pushCallback()`. (Or add the needed helper function into `src/lib/gc.js` if you want!)

    // Set an id to identify each media node - used e.g. by the consoleMedia plugin
    mediaElement = media[i];
    id = mediaElement.getAttribute('id');
    if (id === undefined || id === null) {
      mediaElement.setAttribute('id', 'media-' + type + '-' + i);
      gc.pushCallback( function() { mediaElement.removeAttribute('id'); } );
    }

src/plugins/media/media.js
complanar: Understood :-)

src/plugins/media/media.js
complanar: Adding a new helper function into `src/lib/gc.js` is maybe an idea for a later enhancement of the lib. For the moment your suggestion is enough, to work, thank you.

src/plugins/media/media.js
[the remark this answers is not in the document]
complanar: Okay, you are right with non-surprising default values. But then I would like to have that attributes at the root element and individual steps as well, so that the user is able to overwrite this default behaviour without being forced to add these attributes to each single media.

PR #673, opened 2018-02-24
fix condition in `if` statement

jonschlinkert:
FWIW, it took me a bit longer than it normally would to notice the bug, because of the extra spaces around parens.


henrikingo:
I pushed this fix (in correct source file) and attributed it to you. Thanks for reporting this.

PR #676, opened 2018-02-24
Media

complanar:
Okay, I hope I got this gc stuff right. Can you please review the code again?


src/plugins/media/media.js
```diff
@@ -0,0 +1,166 @@
+/**
+ * Media Plugin
+ * 
+ * This plugin will do the following things:
+ * 
+ *  - Add a special class when playing and pausing media (removing them when ending).
```
henrikingo: More detail: Add a special class `body.impress-media-playing` and `body.impress-media-paused`.

Also document what classes exists on body when presentation has multiple media and some are playing and others are paused.

src/plugins/media/media.js
```diff
@@ -0,0 +1,166 @@
...
+ *
+ *  - Autostart videos when entering a step, if the attribute data-media-autoplay is set and not "false".
+ *
+ *  - Pause all video and audio an the active step when leaving it. This can be disabled for 
+ *    indivudual media nodes by adding the data-media-autopause="false" attribute. 
+ *
```
henrikingo: see previous PR for continued discussion on these.

src/plugins/media/media.js
```diff
@@ -0,0 +1,166 @@
...
+    onStepenter = function (event) {
+        var media, play, i;
+        if ((!event) || (!event.target)) {
+            return;
+        }
+        removeMediaClasses();
+        media = event.target.querySelectorAll('audio, video');
+        for (i = 0; i < media.length; i++) {
+            play = media[i].dataset.mediaAutoplay;
+            if (play !== undefined && play !== "false") {
```
henrikingo: Apologies. On second thought, this should be just:

    if ( play === "true" ) {

Anything else than the string "true" will then be false. This matches a similar attribute in impressConsole.js.

src/plugins/media/media.js
```diff
@@ -0,0 +1,166 @@
...
+                stop = false;
+            }
+            
+            if (pause || stop) {
+                media[i].pause();
+                if (stop) {
+                    media[i].currentTime = 0;
+                }
+            }
+        }
```
henrikingo: I like the general structure here. Previous PR has continued discussion on defaults.

src/plugins/media/media.js
complanar: Sure, but afaik it is common, to skip the true value of boolean values so that
`<video data-media-autoplay>` would be equivalent to `<video data-media-autoplay="true">` just as it is common to type `<video controls>` instead of `<video controls="true">`
Thus we would need 
```javascript
if (play === "" || play === "true")
```
This is compatible to impressConsole *and* html conventions.

src/plugins/media/media.js
complanar: I would only add classes for media on the current step.
Multiple media of the same step being playing or paused at the same time is indeed an issue. Potentially causing head ache for users …

`body.impress-media-playing`, `body.impress-media-paused`, `body.impress-media-id-[ID]-playing` and `body.impress-media-id-[ID]-paused` could be a solution to target specific media. But the body tag is already quite cluttered with class names, so is it really necessary, to add some more?

.gitignore
```diff
@@ -1,3 +1,4 @@
 /js/impress.min.js
 /node_modules
 /npm-debug.log
+/.brackets.json
```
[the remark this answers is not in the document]
complanar: I use Brackets for JS-Development and impress.js with extras contains very many files (over 30,000). Indexing them all slows down Brackets very much, so I hide the extras directory from Bracktes with the help this file, where I set some exclude dirs. 
Of course we do not need it for impress.js, so I added it to `.gitignore`, to prevent git from pushing it to github.

src/plugins/media/media.js
henrikingo: That's a good question :-) I'm just happy to see the autoplay and autostop!

I guess one question is, do you really need classes for paused/stopped? What if there was just a single class `body.impress-media-playing` which is set when at least one media element - perhaps restricted to current step - is playing?

I could also see usefulness in separating audio and video: `impress-media-video-playing` and `impress-media-audio-playing`

src/plugins/media/media.js
henrikingo: I learned HTML during the period where attributes were expected to always have a value (to make it well formed XML), but sure, I'm ok with that.

.gitignore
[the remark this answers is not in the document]
henrikingo: Ok, could you maybe add a comment like this then:

    /js/impress.min.js
    /node_modules
    /npm-debug.log

    +# Files for editors and other tools
    +/.brackets.json

src/plugins/media/media.js
complanar: I like a `*paused` class because I darken the background or fade out other elements when playing videos, but I don’t want them to fade in, when I pause a video for a moment to give some extra explanation. This would not be possible, if there is no class for that. Of course it is up to the user to revert to original styling, when one does not make use of that possibility.

Video and audio are already separated, because maybe one does not want to darken the background when playing audio, but only when playing video (seems logical to me).

complanar: But if you have both `*playing* and `*paused` classes and multiple videos or audios at one step it can have the strange effect, that both classes are present at the same time.
Of course one could remove that other class first (this is the current implementation), but this is maybe not very intuitive behaviour, because in fact you have both states at the same time.

henrikingo: So if you have: `impress-media-video-playing, impress-media-video-paused` (and same for audio), but no `-stopped` class, then you end up with:

* -playing = at least one video (or audio) is playing
* -paused = at least one video (audio) is paused
* Nothing = no video (audio) exists, or if it does, it is in stopped state

Those would be clear to me. Note that authoring meaningful CSS for the case where multiple videos are playing or paused simultaneously may not be easy or possible. IMO that's ok. An author can also write their own javascript to catch events and do whatever they want, if it's no possible from above menu.

src/plugins/media/media.js
```diff
@@ -0,0 +1,261 @@
...
+        delete root.mediaAutoplay;
+        delete root.mediaAutopause;
+        delete root.mediaAutopstop;
+        elements = document.querySelectorAll("#root, .step, audio, video");
+        for (i = 0; i < elements.length; i += 1) {
+            el = elements[i];
+            delete el.mediaAutoplay;
+            delete el.mediaAutopause;
+            delete el.mediaAutostop;
+        }
```
henrikingo: The above attributes are created by the user. You should not delete them.

The below are automatically created by yourself, those you should delete.

In other words, you can delete the above code block.

src/plugins/media/media.js
```diff
@@ -0,0 +1,261 @@
...
+            root.mediaAutopause = false;
+        }
+        
+        // this *must* be called only after setting the autopause, autoplay and autostop values
+        enhanceMedia();
+        
+        gc.pushCallback(garbage);
+    }, false);
+    
+    garbage = function () {
```
henrikingo: gc/garbageCollector is not a perfect name, as it's not really garbage and not really a garbage collector.

To not perpetuate my poor choice of naming here, could you call this function something like `cleanup`, `cleanupAttributes` or `teardown` or `teardownCallback`.

src/plugins/media/media.js
```diff
@@ -0,0 +1,261 @@
...
+ *
+ *    - data-media-autostart="true" data-media-autostop="true" data-media-autopause="true": start 
+ *      media on enter, stop on leave (stop overwrites pause).
+ *
+ *    - <div id="impress" data-media-autostart="true"> ... <div class="step" data-media-autostart="false">
+ *      All media is startet automatically on all steps except the one that has the data-media-autostart="false"
+ *      attribute.
+ *
+ *  - Pro tip: Use <audio onended="impress().next()"> or <video onended="impress().next()"> to proceed to the 
+ *    next step automatically, when the end of the media is reached.
```
henrikingo: Very good docs!

src/plugins/media/media.js
```diff
@@ -0,0 +1,261 @@
...
+        for (type in types) {
+            document.body.classList.remove("impress-media-" + types[type] + "-playing");
+            document.body.classList.remove("impress-media-" + types[type] + "-paused");
+        }
+    };
+
+    enhanceMediaNodes = function () {
+        var i, id, media, mediaElement, type;
+        
+        //gc = event.detail.api.lib.gc;
```
henrikingo: Above can be removed.

src/plugins/media/media.js
```diff
@@ -0,0 +1,261 @@
...
+
+    enhanceMedia = function () {
+        var steps, stepElement, i;
+        enhanceMediaNodes();
+        steps = document.getElementsByClassName("step");
+        for (i = 0; i < steps.length; i += 1) {
+            stepElement = steps[i];
+            
+            // Inherit autoplay, autostop and autopause settings from root element if there is no own setting
+            if (stepElement.dataset.mediaAutoplay === undefined || stepElement.dataset.mediaAutoplay === null) {
```
henrikingo: Ah. The below approach is straightforward, but makes it impossible to do the teardown. After you populate default values into the actual DOM objects themselves, you lose track of whether they were created by the user or by this code.

My preferred solution is that you don't do this at all in the init phase, and instead implement a function that returns the right value (combination of mediaElement.dataset.media*, stepElement.dataset.media*, rootElement.dataset.media* and the default) for a given media element.

src/plugins/media/media.js
```diff
@@ -0,0 +1,261 @@
...
+        if ((!event) || (!event.target)) {
+            return;
+        }
+        removeMediaClasses();
+        stepElement = event.target;
+        media = event.target.querySelectorAll("audio, video");
+        for (i = 0; i < media.length; i += 1) {
+            mediaElement = media[i];
+            
+            // Inherit autoplay settings from step element if there is no own setting
```
henrikingo: Same problem here.

src/plugins/media/media.js
```diff
@@ -0,0 +1,261 @@
...
+            if (mediaElement.dataset.mediaAutoplay === undefined || mediaElement.dataset.mediaAutoplay === null) {
+                mediaElement.mediaAutoplay = stepElement.mediaAutoplay;
+            } else {
+                mediaElement.mediaAutoplay = mediaElement.dataset.mediaAutoplay === "" || mediaElement.dataset.mediaAutoplay === "true";
+            }
+            
+            // Autoplay when true, but only if not on preview of the next step in impressConsole
+            onConsolePreview = (window.frameElement !== null && window.frameElement.id === "preView");
+            if (mediaElement.mediaAutoplay && !onConsolePreview) {
```
henrikingo: So, this would now become `getAttrValue('autoplay', mediaElement)`

src/plugins/media/media.js
complanar: Good suggestion, thank you.

src/plugins/media/media.js
complanar: Your comment confuses me. Are custom properties I add to a DOM object automatically transferred into HTML attributes?

`el.dataset.mediaAutoplay` refers to the user defined HTML attributes
`el.mediaAutoplay` refers to my defined properties.

This should be comparable to the way the rel plugin works, isn’t it? Or am I missing anything important here?

src/plugins/media/media.js
complanar: Sorry, forgot to remove the comment. Seems to be an artifact of the development process.

src/plugins/media/media.js
complanar: See my above comment to lines 100 to 103. (BTW is there any way in github markdown to refer to some comment in a way that turns into a link?)

Implementing a function to get this value at runtime is indeed a solution. My thought was that caching them as properties of an object would maybe speed up the move to the next step. Doing the same work over and over again (for maybe a lot of audio objects at one single step) should be slower than calculating it once.

Of course this would make a dynamic change of those attributes at runtime easier, but I don’t consider this a common scenario.

src/plugins/media/media.js
```diff
@@ -0,0 +1,261 @@
...
+            gc.addEventListener(media[i], "playing", onPlay);
+            gc.addEventListener(media[i], "pause", onPause);
+            gc.addEventListener(media[i], "ended", onEnded);
+        }
+    };
+
+    enhanceMedia = function () {
+        var steps, stepElement, i;
+        enhanceMediaNodes();
+        steps = document.getElementsByClassName("step");
```
henrikingo: I have missed this previously: This needs to be `root.getElements...`

src/plugins/media/media.js
henrikingo: Ah right. I was indeed reading you wrong.

The `rel` plugin keeps these in an internal data structure `startingState`. It's a list of plain JavaScript objects, not DOM. (impress.js core does the same for its list of steps and their positions.) It's not a big difference, as that internal list too has to be reset at teardown. Even then, stylistically I'd prefer you stay with the same approach here.

> Implementing a function to get this value at runtime is indeed a solution. My thought was that caching them as properties of an object would maybe speed up the move to the next step.

You're correct, of course. In this case I prefer simpler, with no caching. It's just 3 lookups for a given element, so keeping the code simpler is a benefit too.

> Of course this would make a dynamic change of those attributes at runtime easier, but I don’t consider this a common scenario.

The use case where that happens is impressionist, the wysiwyg editor project. That's where the need to use gc comes from too. While impress.js is its own standalone project, I try to keep in mind that it should work also when embedded into something like impressionist. (Without adding too much complexity or other hardship, of course.)

henrikingo: > (BTW is there any way in github markdown to refer to some comment in a way that turns into a link?)

I think the date, such as "15 hours ago" is such a link.

But you don't need to, I of course know what you were referencing. You can also reference the line number where the comment is at.

src/plugins/media/media.js
```diff
@@ -99,70 +169,74 @@
...
-        for (i = 0; i < media.length; i++) {
-            play = media[i].dataset.mediaAutoplay;
-            if (play !== undefined && play !== "false") {
-                media[i].play();
+        
+        media = stepElement.querySelectorAll("audio, video");
+        for (i = 0; i < media.length; i += 1) {
+            mediaElement = media[i];
+            
+            // Autoplay when (maybe inherited) autoplay setting is true,
```
henrikingo: Could you try to move all impressConsole related stuff to a separate helper function?

* The onConsoleSlideView part you should be able to handle completely inside that function
* The function could return true if we are in fact in any window that is part of impressConsole

This would simplify the code that remains here to something like:

    mediaElement = media[i]
    isConsole = onStepenterImpressConsole(mediaElement);
    if (getMediaAttribute(...) && !isConsole) {
        mediaElement.play()
    }

src/plugins/media/media.js
henrikingo: Noting to myself I have replied to this in another thread.

src/plugins/media/media.js
```diff
@@ -36,60 +59,107 @@
...
+            delete el.mediaAutostop;
+        }
+        for (i = 0; i < attributeTracker.length; i += 1) {
+            el = attributeTracker[i];
+            el.node.removeAttribute(el.attr);
+        }
+        attributeTracker = [];
+    };
+    
+    getMediaAttribute = function (attributeName, nodeList) {
```
henrikingo: Could you avoid the name "nodeList". In the DOM there's a class "NodeList" and this is not of that type, so it would be better to use any other neutral name. Like just "nodes" or "nodeHierarchy".

src/plugins/media/media.js
```diff
@@ -99,70 +169,74 @@
...
-            }
-            
-            if (pause || stop) {
-                media[i].pause();
-                if (stop) {
-                    media[i].currentTime = 0;
+            if (autopause || autostop) {
+                mediaElement.pause();
+                if (autostop) {
+                    mediaElement.currentTime = 0;
```
henrikingo: Now even nicer than before.

henrikingo:
Note btw that the repo has testing and linting enabled. You don't really affect existing tests, but you need to pass linting: `npm run lint`. The above check executes the commands from /circle.yml

complanar:
I already noted that failing checks, but all complaints were about code styling errors, so I did not remove them.
BTW I  find that code style a quite strange, and my own linter complains about that style heavily ... Maybe it’s just a matter of taste or habit.

henrikingo:
I was not part of the decision, but if I remember correctly, the jquery style was selected and then we just follow that. (At least for one of the linters.)

src/plugins/media/media.js
```diff
@@ -0,0 +1,248 @@
...
+            mediaElement = media[ i ];
+
+            // Autoplay when (maybe inherited) autoplay setting is true,
+            // but only if not on preview of the next step in impressConsole
+            onConsole = onStepenterDetectImpressConsole();
+            autoplay = getMediaAttribute( "autoplay", [ mediaElement, stepElement, root ] );
+            if ( autoplay && !onConsole.preview ) {
+                if ( onConsole.slideView ) {
+                    mediaElement.muted = true;
+                }
```
henrikingo: Lines 210-212 can be moved inside onStepenterDetectImpressConsole(). (Line 209 can stay as is.)

complanar: I am not sure about this. That would mute media at the console view even when it is not played automatically. 

I don’t think we should mute every media inside console view even when it maybe is not started. Chances are high, that somebody has an audio on a step and starts them inside the console. Then can not be heard at all, because it is (surprisingly) muted. Audio has not necessarily be played on both screens (or even the main) in contradiction to video that should always be played on both screens, at least inside the presentation view.

I think this should only be done by an additional plugin that handles events between main and console view (we already talked about that). I already had a script working for bi-directional synchronization of media, even avoiding circular propagation from one window to the other, but thought it best to implement this as separate plugin as it assumes the existence of the impressConsole plugin and their corresponding windows. (I am not sure if the current design of the console, especially the relation between main and slave window is a good one, either. What about interchanging their roles as in most cases users would expect the presentation view to respond to their actions inside the console window but not the other way round, even if this could sometimes be handy. But maybe we should discuss this separately.)

It would be good to have some kind of general plugin for that kind of synchronization, not only for media but  also for forms, hover/highlight/pointer events and so on, but I did not have an idea how to accomplish that without catching every single event by hand, filtering unwanted ones and propagating the interesting ones to the other window.

I would let it be the way it is at the moment.

complanar: Maybe one could implement this console detection more generally in lib because it could be handy for some other plugins.
In this case the function should not trigger *any* actions.

henrikingo: > That would mute media at the console view even when it is not played automatically.

Ah, good point. I haven't considered that use case.

You could still achieve the same goal while taking that into account, but I think my suggestion has in any case lost its elegance at that point and if anything your approach is now clearer.

> In this case the function should not trigger any actions.

I agree.

complanar:
Yes, I think so.

PR #678, opened 2018-03-06
Polar

lwhshugyosha:
polar plugin reworked for separate single-step positioning.
Only polar.js file changed, other changes not to be pulled


src/plugins/polar/README.md
```diff
@@ -0,0 +1,96 @@
...
+-----------------
+
+To place steps e.g. in a circle around a central thought, polar positioning is introduced here.
+It uses the following attributes for a step:
+	data-rel-ref
+	data-rho
+	data-phi
+	data-rel-rho
+	data-rel-phi
+
```
complanar: Why do you use rho and phi for polar coordinates? As rho is a greek letter it is associated most likely with an angle not a length. I think most rho and phi are common for spherical coordinates, as you have one more dimension. 
I would suggest `-distance` and `-phi` for polar coordinates and `-distance`, `-rho` and `-phi` or `-primary-angle` and `-secondary-angle` for spherical coordinates.
Have a look at https://en.wikipedia.org/wiki/Spherical_coordinate_system and https://en.wikipedia.org/wiki/Spherical_coordinate_system.

src/plugins/polar/README.md
```diff
@@ -0,0 +1,96 @@
...
+Rho is the radius (the distance from the origin) and phi the angle in degree, counting from the 
+horizontal (positive x, "03:00hrs") counterclockwise.
+In addition to plain numbers, which are pixel values, it is also possible to define the radius 
+as a multiple of the normal step diagonal, using a unit of "d" appended to the number.
+
+Relative polar coordinates are not added up, they just count from the given origin. For absolute
+coordinates that is (0,0,0), for relative ones the origin of the last step or the one pointed to 
+by the "data-rel-ref" attribute. 
+Alternatively, a new relative origin can be set by giving absolute cartesian coordinates.
+
```
complanar: What if no `data-rel-ref` attribute is given, but `data-rel-rho` and `data-rel-phi` attributes are present?
Doesn't that mean, that the coordinates are meant relative to the previous step? In such a way the would kind of "add up", if multiple steps are defined that way, at least I would expect that.

src/plugins/polar/README.md
lwhshugyosha: Forces of habit. I learned to use rho, phi and theta forty years ago and still do ;-)

src/plugins/polar/README.md
lwhshugyosha: Sorry, Must be `data-pos-ref`. This is only used for 'absolute' coordinates. 
Relative coordinates are always from the last step and are inherited and add up until some other positioning is defined. This is the same as I understand the original rel plugin is supposed to work.

And yes, it does the same as the rel plugin, as all positioning must be done for every step at the same time - otherwise no reference would work if a later plugin relocates the referred step. Therefore it's either the rel plugin or this one.

src/plugins/polar/README.md
lwhshugyosha: BTW: x, y and z are also not 'quite' standard, here they are left-handed to comply with screen definitions...

henrikingo:
> Only polar.js file changed, other changes not to be pulled

Sorry but this is too messy to look at. When you're done with your presentation / whatever you're doing now, please submit a new PR with just the code actually intended for merging, so we can focus on that. I suggest you close this PR until then, but I don't mind if it's kept open, if you have some ongoing discussion.

PR #683, opened 2018-03-18
Rel

complanar:
Add relative positioning to any previous step.


src/plugins/rel/rel.js
```diff
@@ -86,6 +86,36 @@
             prev = { x:0, y:0, z:0, relative: { x:0, y:0, z:0 } };
         }
 
+        if ( data.relTo ) {
+
+            var ref = root.getElementById( data.relTo );
+            if ( ref ) {
+
+                // Test, if it is a previous step that already has some assigned position data
+                if ( el.compareDocumentPosition( ref ) & Node.DOCUMENT_POSITION_PRECEDING ) {
```
henrikingo: Cool. I've never done that before.

src/plugins/rel/rel.js
```diff
@@ -86,6 +86,36 @@
...
+                    prev.x = ref.getAttribute( "data-x" );
+                    prev.y = ref.getAttribute( "data-y" );
+                    prev.z = ref.getAttribute( "data-z" );
```
henrikingo: It has no functional effect in current code, but to not leave old data dangling here, can you also add this line:

    prev.relative = {};

PR #684, opened 2018-03-18
Spherical

complanar:
Add spherical and polar positioning (polar if `data-rel-theta` attribute is ommited).


complanar:
Sorry, ran into merge problems, closing for now and creating a new PR.

PR #685, opened 2018-03-18
Spherical

complanar:
Add spherical positioning plugin.


src/plugins/spherical/spherical.js
```diff
@@ -0,0 +1,147 @@
...
+                el.setAttribute( "data-spherical-theta", 90 );
+            }
+
+            cartesian = sphericalToCartesian(
+                toNumber( data.sphericalDistance, 0 ),
+                toNumber( data.sphericalPhi * Math.PI / 180, 0 ),
+                toNumber( data.sphericalTheta * Math.PI / 180, 90 )
+            );
+
+            if ( data.x === undefined ) {
```
henrikingo: I don't think this works. Use of spherical coordinates will generally always set all of x/y/z. Therefore you cannot check and set them one by one, rather it is all or nothing:

* If any data-spherical-* attributes are set, then there must be no data-rel-x/y/z and no data-x/y/z attributes set. 
* If even one of them is set, then all data-spherical-* attributes are ignored and a warning logged.

src/plugins/spherical/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+Following html attributes are supported for step elements:
+
+    data-spherical-dist
+    data-spherical-rho
+    data-spherical-theta
+    
+    data-spherical-rel-dist
+    data-sphercial-rel-rho
+    data-spherical-rel-theta
```
henrikingo: I think this can be simplified. IMO we should think of spherical coordinates as always being relative to something. Therefore you don't need the separate `data-spherical-rel-*` group of attributes. Rather, if user specified `data-rel-to`, then treat `data-spherical-*` attributes as relative to that, and if there is no `data-rel-to` then default to just setting `data-x/y/z` directly against origin.

complanar: Then we would need a special "keyword" that indicates that a step should be located relative to the previous step or we would loose the functionality that a step can be positioned relative to its predecessor without explicitly defining a reference, wouldn't we?

henrikingo: My suggestion is that for spherical / polar coordinates, it would be an unlikely use case that each slide would be relative to its predecessor. It seems to me it's much more likely they will all be relative to a common point - either the zero origin, or a common reference step.

Of course, if you wanted to make spherical steps relative to their predecessor, then each would in fact have to explicitly say so with `data-rel-to`. So it's a tradeoff, trying to make a smooth experience for the common use case.

Do you disagree with this assumption of the common use case.

complanar: What about treating them as relative, when `data-rel-to` is present or if the `*-rel-*` versions are used. This rule could be applied to the rel plugin, too. Of course this would need a different approach of setting positioning data as it changes the meaning of `data-x`, `data-y` and `data-z` depending on the existence (and validity) of other attributes.
I was already thinking of a completely new positioning system à la tikz/pgf (a latex package for creating graphs and other vector based pictures). There is only one positioning attribute that follows a simple parsed micro syntax.

-    `(x,y)` or `(x,y,z)` absolute cartesian values
-    `(angle:distance)` or `(angle:angle:distance)` polar and spherical coordinates
-    `+(x,y)` or `+(x,y,z)` or `+(angle:distance)` relative cartesian or polar values
-    `(x1,y1,z1)+(x2,y2,z2)` addition of two coordinates
-    `(ref)` reference to another position (here: step)
-    `($expression$)` evaluated expression, e.g. ...
-    `($(x,y,z)+(angle:angle:distance)$)` combining adding cartesian and polar coordinates
-    `($(ref)+(x,y,z)$)` adding a value to a reference position, similar to data-rel-to
-    `($(ref1)!0.5!(ref2)$)` in the middle between two reference positions
-    `($(ref1)!0.2!(ref2)$)` as above but closer to ref1
-    `($(ref1)!1.5!(ref2)$)` along the vector from ref1 to ref2, 1.5 times away the distance from ref1 to ref2
-    this could be extended not only to use pixel lengths but arbitrary css units. Conversion between them and mixing them is of course a complicated story ...

I think this is far beyond the scope of this PR but I use tikz/pgf very often and I like its flexible coordinate system (much more possibilities than I showed here) very much.

complanar: Oh, I forgot to say: I could live with your assumption of common use cases :-)

PR #702, opened 2018-08-27
Fix some touch bugs on mobile phones.

14725:
Call event.preventDefault() to prevent scrolling on WebKits and to prevent strange "touchcancel" events on Android WebViews.

### DIFF
--- a/src/plugins/touch/touch.js
+++ b/src/plugins/touch/touch.js
@@ -25,6 +25,7 @@
 
     document.addEventListener( "touchstart", function( event ) {
         lastX = startX = event.touches[ 0 ].clientX;
+        event.preventDefault();
     } );
 
     document.addEventListener( "touchmove", function( event ) {
@@ -36,6 +37,7 @@
          lastX = x;
 
          window.impress().swipe( diff / window.innerWidth );
+         event.preventDefault();
      } );
 
      document.addEventListener( "touchend", function() {


src/plugins/touch/touch.js
```diff
@@ -25,6 +25,7 @@
 
     document.addEventListener( "touchstart", function( event ) {
         lastX = startX = event.touches[ 0 ].clientX;
+        event.preventDefault();
```
henrikingo: Is this one necessary? Seems like it could prevent legitimate touchstart events reaching their intended destinations.

If it is necessary, please add a comment to explain exactly what problem this fixes, on which browsers, etc.

src/plugins/touch/touch.js
```diff
@@ -36,6 +37,7 @@
          lastX = x;
 
          window.impress().swipe( diff / window.innerWidth );
+         event.preventDefault();
```
henrikingo: This one seems more defensible, since it is intended that any swipe = touchmove should end up in this plugin and nowhere else.

src/plugins/touch/touch.js
14725: In fact, I know nothing about it.
Well, I changed some code in my copy to make it works on old browsers. (etg. for of -> for in) I test it on Via (an browser using Android's WebView, for old phones) on Android 4.4, but I can't swipe. 
It is said that it will auto 'touchcancel' after 'touchstart' with scrolling.So I add this. 
( https://m.2cto.com/kf/201411/348858.html )...

14725: Maybe new browsers needn't event.preventDefault(); I haven't tested yet.

Issue #31, opened 2012-01-05
Scroll!


fj1800:
First - extremely impressed by your work! Though, the arrows might not be perfectly intuitive for everyone. Clicking neither. Adding Scroll to the ways of navigating would be a huge plus, and it shouldn't be too difficult using Scrolldeck (http://johnpolacek.github.com/scrolldeck.js/) or something like that.

vladimirlisovets:
I spent the whole day trying to implement on web dev. diverse presentation plugins, including impress.
Impress is the best to upgrade to such purpose.
Sadly, I didn't find anything, hope someday, talented impress developers would find a way to offer such option.

xprilion:
Hey! I have been able to implement it! I will shortly create a fork with that implemented :smiley:

PR #704, opened 2018-10-01
Added Scroll example

xprilion:
Added an example to allow scrolling through slides using mouse/touchpad. This resolves issue https://github.com/impress/impress.js/issues/31


henrikingo:
Hi @xprilion !

Thanks for your contribution! I think the idea and code are roughly ok, but you need to make the code a plugin: https://github.com/impress/impress.js/blob/master/src/plugins/README.md

For name of the plugin I suggest: `navigation-wheel`

Also, instead of waiting for a fixed 1 second, it might fit better if you wait for the time set in data-transition-duration. (This is often 1 second, but user configurable.)

Let me know if you need more help. I will close this PR now, but you can still use this thread for discussion. Please open a fresh PR if you want to submit the same code as a plugin.

Issue #220, opened 2012-11-09
'?' key should pop up keyboard shortcut help window


aspiers:
'?' is being rapidly adopted as the standard keyboard shortcut for AJAX-ified web apps to pop up a help window listing useful information such as which shortcuts are supported for keyboard navigation etc.  Some examples:
- google mail
- google reader
- google calendar
- google news
- github
- Trello

It would be great if impress.js did this.  The help window could of course also link to the impress.js documentation and github site, which would help grow your community ;-)

FagnerMartinsBrack:
Hi @aspiers , in an effort to clear up older issues/PRs we are pinging back to know if you are still tracking this request.

To give a little bit of context, recently [a decision was made in the project](https://github.com/impress/impress.js/issues/435#issuecomment-177602861) to make the development more active and the first task is to clear up older issues like this one to see if [the OP](http://netforbeginners.about.com/od/internetlanguage/f/What-Is-OP.htm) is still interested in keep it going.

Do you have any links or examples that show if this standard still exists?

aspiers:
Yes sure - I'm very much interested in seeing this implemented.  Some obvious examples:

- GitHub
- Gmail
- Google Calendar
- Google Drive
- Trello
- gerrit

but I am sure there are many others following the example of these very popular sites.

aspiers:
Oh sorry, I just noticed I already gave a list in the original issue description :-)

FagnerMartinsBrack:
> Oh sorry, I just noticed I already gave a list in the original issue description :-)

Yes I saw the list in the description, what I was asking was a demonstration of the feature. I want to verify how it exactly works to implement correctly if this is accepted and if the feature still exists in the services you mentioned. A link or a video using LICECap demonstrating the feature would be enough. Thank you!

aspiers:
@FagnerMartinsBrack I think maybe you are fundamentally misunderstanding the feature request? Because you seem to think it's way more complicated than it actually is:

1. visit one of the sites I mentioned
2. press the '?' key
3. observe a popup window listing all the keyboard shortcuts available

You definitely don't need any videos for this!  It takes about 10 seconds per site to verify that they do it.

FagnerMartinsBrack:
> I think maybe you are fundamentally misunderstanding the feature request? 

Yeah, I definitely misunderstood. Somehow I interpreted as a `?` in the querystring or something like that. Personally I never knew such feature existed consistently between services like these. Thanks :)

aspiers:
[...]

PR #705, opened 2018-10-07
Help plugin pops up with ?

zilioti:
As described in #220, I have just added another key press to open plugin help.


henrikingo:
Thanks!

PR #707, opened 2018-10-29
Add relative rotation to rel plugin

epsilonhalbe:
Adds: new keywords `data-rel-rotate`, `data-rel-rotate-x`, `data-rel-rotate-y`, and `data-rel-rotate-z`

NOTE: This breaks the continuously progressing `data-rel-x` feature, by resetting the value to 0, if missing, I don't think it is too much trouble to add the "data-rel-x=1000" every slide and makes things much clearer.

### DIFF
--- a/src/plugins/rel/rel.js
+++ b/src/plugins/rel/rel.js
@@ -10,21 +10,34 @@
  * Example:
  *
  *         <!-- Position step 1000 px to the right and 500 px up from the previous step. -->
- *         <div class="step" data-rel-x="1000" data-rel-y="500">
+ *         <div class="step" data-rel-x="1000" data-rel-y="500" data-rel-rotate="90" >
+ *
+ *         <!-- Position step 1000 px to the left and 750 px up from the step with id "title". -->
+ *         <div class="step" data-rel-x="-1000" data-rel-y="750" data-rel-to="title">
  *
  * Following html attributes are supported for step elements:
+ *
+ *     data-rel-to
  *
  *     data-rel-x
  *     data-rel-y
  *     data-rel-z
  *
- * These values are also inherited from the previous step. This makes it easy to
- * create a boring presentation where each slide shifts for example 1000px down
- * from the previous.
+ *     data-rel-rotate-x
+ *     data-rel-rotate-y
+ *     data-rel-rotate-z
+ *     // or equivalently
+ *     data-rel-rotate
  *
- * In addition to plain numbers, which are pixel values, it is also possible to
- * define relative positions as a multiple of screen height and width, using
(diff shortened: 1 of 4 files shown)


src/plugins/rel/rel.js
```diff
@@ -196,6 +213,23 @@
...
+                    step.el.removeAttribute( "data-rotate-y" );
+                } else {
+                    step.el.setAttribute( "data-rotate-y", step.rotateY );
+                }
+                if ( step.rotateZ === null ) {
+                    step.el.removeAttribute( "data-rotate-z" );
+                    step.el.removeAttribute( "data-rotate" );
+                } else {
+                    step.el.setAttribute( "data-rotate-z", step.rotateZ );
+                    step.el.setAttribute( "data-rotate", step.rotateZ );
```
henrikingo: The point here is to return the initial state. You can't set both data-rotate and data-rotate-z, rather you must set them exactly as they were when the html was loaded. (Or rather, at the time that impress().init() was called.)

src/plugins/rel/rel.js
```diff
@@ -159,14 +164,26 @@
...
             el.setAttribute( "data-z", step.z );
+            el.setAttribute( "data-rotate-x", step.rotateX );
+            el.setAttribute( "data-rotate-y", step.rotateY );
+            el.setAttribute( "data-rotate-z", step.rotateZ );
+            el.removeAttribute( "data-rel-x" );
+            el.removeAttribute( "data-rel-y" );
+            el.removeAttribute( "data-rel-z" );
+            el.removeAttribute( "data-rel-rotate-x" );
+            el.removeAttribute( "data-rel-rotate-y" );
+            el.removeAttribute( "data-rel-rotate-z" );
```
henrikingo: Why are the removeAttribute()s needed? Note that if you want to remove some attribute, you must reset the original value in the lib.gc.pushCallback() function.

src/plugins/rel/rel.js
```diff
@@ -117,36 +133,25 @@
...
-        step.y = step.y + step.relative.y;
-        step.z = step.z + step.relative.z;
-
-        return step;
+        return {
+
+          // If data-? is set, discard any data-rel-? values otherwise use data-rel-?
+          // If neither data-? nor data-rel-? are set - use 0 as a default value
+          x: data.x !== undefined ? toNumber( data.x ) :
+              ( data.relX === undefined ? 0 : prev.x + toNumberAdvanced( data.relX ) ),
```
henrikingo: The point of using toNumber() is to keep this block of code simple and easy to read. If you need to do these complex comparisons, create a new function that does all of these nicely:

    x: figureOutCoordinate( ... ),

Issue #681, opened 2018-03-09
Introduce optional but standard css file for plugin related css


henrikingo:
impress.js doesn't require any particular CSS file. Each author should create their own, to achieve the visual style they want.

In practice, as we've added many plugins, a plugin will not do anything useful without CSS. (See for example mouse-timeout plugin.) The src/plugins/README.md provides example CSS that a user may want to copy paste.

It's significant enough, that it makes sense to move that CSS out into an actual css file, like css/impress-common.css. After that, all the demo presentations should be also changed to use that CSS file, and the same CSS should be removed from each of their own CSS.

PR #714, opened 2019-01-15
Add full help screen to help plugin

giflw:
All actual help still appears on short help.
Use parameter `short: true` to show message on short and full help.

### DIFF
--- a/src/plugins/help/help.js
+++ b/src/plugins/help/help.js
@@ -1,11 +1,6 @@
 /**
  * Help popup plugin
  *
- * Example:
- *
- *     <!-- Show a help popup at start, or if user presses "H" -->
- *     <div id="impress-help"></div>
- *
  * For developers:
  *
  * Typical use for this plugin, is for plugins that support some keypress, to add a line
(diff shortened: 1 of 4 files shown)


src/plugins/help/help.js
```diff
@@ -1,11 +1,6 @@
 /**
  * Help popup plugin
  *
- * Example:
- *
- *     <!-- Show a help popup at start, or if user presses "H" -->
- *     <div id="impress-help"></div>
- *
```
henrikingo: This needs to stay. Plugins that display something on the screen are required to be optional. One way to make them optional is that the presentation author must explicitly add an empty div for them to appear.

In the case of the help plugin, note that it is shown automatically when the presentation is loaded. Due to this, the div must stay.

src/plugins/help/help.js
```diff
@@ -21,46 +16,88 @@
...
+        position: "fixed",
+        top: 0,
+        bottom: 0,
+        left: 0,
+        right: 0,
+        backgroundColor: "#000",
+        color: "#FFF",
+        opacity: 0.8,
+        textAlign: "center",
+        padding: "3em"
```
henrikingo: The philosophy of impress.js is that we don't want to force any colors or other styling for the user. So instead of setting the style here, you should set it in css/impress-demo.css. Unfortunately this means you also need to set this in every other css file under presentations. (Related: #681)

So what is left for the plugin to do? Maybe nothing? Do you need different CSS for the full help?

At most, you could set a class that indicates short help vs full help:

    <div id="impress-help" class="help-short">
    <div id="impress-help" class="help-full">

Presentation authors can then choose to use this class for styling if they want.

simplePCBuilding:
I don't think this is something that needs a plugin update. I would much rather have it in a impress-extra.css file. I also think that we should have like a impress.css file where important and common CSS is defined, which may or may not be included

PR #716, opened 2019-01-15
Add support for "." to enter/exit blackout screen

giflw:
- This is the default on Power Point
- THis add support for remote controller presentation blackout key


giflw:
@henrikingo Done renaming resume to play on autoplay plugin

PR #717, opened 2019-01-15
Add "o" as overview navigation

giflw:
This shows the overview slide, if provided
<div id="impress-overview">...</div>

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -3140,8 +3140,18 @@
 ( function( document ) {
     "use strict";
 
+    var overviewId = "impress-overview";
+
     // Wait for impress.js to be initialized
     document.addEventListener( "impress:init", function( event ) {
+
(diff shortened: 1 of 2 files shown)


henrikingo:
Ok, so this one I would like to approach completely differently. Instead of thinking of the overview slide as a special case, let's do something that achieves the same but in a more general purpose solution.

Create a completely new plugin, which we may call `accesskey` based on the similarly named HTML attribute. Then, modify the overview step like this:

    <div id="overview" class="step" data-accesskey="o" data-x="..." data-y="...">

Then implement accesskey.js so that I can bind any letter to any slide I want.

You can, and should, modify the index.html demo and the presentations under examples so that their overview slide uses this feature.

henrikingo:
I believe #852 implements this

PR #721, opened 2019-02-24
Update dependencies and remove outdated ones

mohe2015:
This also fixes some vulnerabilities in dependencies (shouldn't be an issue because code is only executed locally)
At some time the eslint fixes should be applied (especially reformatting, or the config should be changed)
Maybe package-lock.json should also be committed as [recommended](https://docs.npmjs.com/files/package-lock.json)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,24 @@
+module.exports = {
+    "env": {
+        "browser": true,
+        "node": true
+    },
+    "extends": "eslint:recommended",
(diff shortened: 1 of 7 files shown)


mohe2015:
This has multiple issues - I tested with a "too new" nodejs version, ...

henrikingo:
Shame. I did like this PR. Thanks for the effort at least, I do appreciate this kind of boring cleanup work.

PR #728, opened 2019-04-13
swiping up and down triggers arrow down and up

jhepp:
Slide your finger on a touch device up will trigger the "arrow down"-key and slide down will trigger the "arrow up"-key.

An example how to use it:
`data-goto-key-list="ArrowUp ArrowDown ArrowLeft ArrowRight"`
`data-goto-next-list="myup mydown myleft myright"`

### DIFF
--- a/src/plugins/touch/touch.js
+++ b/src/plugins/touch/touch.js
@@ -15,43 +15,81 @@
 ( function( document, window ) {
     "use strict";
 
-    // Touch handler to detect swiping left and right based on window size.
-    // If the difference in X change is bigger than 1/20 of the screen width,
-    // we simply call an appropriate API function to complete the transition.
+    // Touch handler to detect swiping left and right, up and down based on
+    // window size.
+    // If the difference in X or Y change is bigger than 1/20 of the screen
+    // width or height, we simply call an appropriate API function to complete
+    // the transition.
     var startX = 0;
+    var startY = 0;
     var lastX = 0;
+    var lastY = 0;
     var lastDX = 0;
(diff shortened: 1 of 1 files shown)


henrikingo:
Thanks Jochen!

Code looks ok.

What was your motivation for adding this? Is it to be used together with goto-key-list, like in https://impress.js.org/examples/2D-navigation/#/step-1 ?

The only thing I'm wondering about is that now left-right swipes call next() and prev() explicitly, but up-down is translated to arrow keys. Maybe it would be more useful, and consistent, if you changed also left-right to send arrow keys? What do you think?

jhepp:
The motivation for adding this code was to be able on a touch only device with no accessible keyboard to use a goto-key-list and goto-next-list navigation like in the example you mentioned.

Swiping up and down offers a second navigation dimension to touch only devices.

The translation of up-down swipes to arrow keys is a detour because i found no solution to call "up()" and "down()" directly.

next() and prev() are part of the ImpressAPI whereas "up()" and "down()" are handled by the navigation plugin. So I decided not to not change this.

henrikingo:
Thanks - and sorry for my absence due to the traditional spring flu!

Could you make the change so that also left and right swipes map to arrow keys? For example, if you try the examples/2D-navigation in your branch now, it doesn't navigate in a meaningful way, because "next" in the presentation flow is actually down. So I believe the right solution is to map all swipe directions to arrow keys. This will allow presentation authors to override all of them with goto, and by default the arrows will trigger next/prev just like they do from a keyboard. (So both right and down are next.)

PR #729, opened 2019-04-13
skip autoplay for some time after user interaction

jhepp:
Skip autoplay if a key press, a key release, a mouse move, a mouse button click or a touch is noticed.

To skip two autoplays after a user interaction use:
`data-user-skip-autoplay="2"`

### DIFF
--- a/src/plugins/autoplay/autoplay.js
+++ b/src/plugins/autoplay/autoplay.js
@@ -10,6 +10,8 @@
     "use strict";
 
     var autoplayDefault = 0;
+    var userSkipAutoplayDefault = 0;
+    var userSkipAutoplay = 0;
     var currentStepTimeout = 0;
     var api = null;
     var timeoutHandle = null;
@@ -26,6 +28,7 @@
         // or anything. `impress:init` event data gives you everything you
         // need to control the presentation that was just initialized.
         api = event.detail.api;
+        var gc = api.lib.gc;
         root = event.target;
 
         // Element attributes starting with "data-", become available under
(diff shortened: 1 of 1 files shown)


src/plugins/autoplay/autoplay.js
```diff
@@ -59,6 +72,15 @@
         reloadTimeout( event );
     }, false );
 
+    var user = function( event ) {
```
henrikingo: Can you call this `userActivity`

src/plugins/autoplay/autoplay.js
```diff
@@ -10,6 +10,8 @@
     "use strict";
 
     var autoplayDefault = 0;
+    var userSkipAutoplayDefault = 0;
```
henrikingo: It looks like I was lazy and there's no documentation for this plugin. (Either in the comment above, or a separate README file.) Could you add it now? Including your own addition.

src/plugins/autoplay/autoplay.js
```diff
@@ -36,15 +39,25 @@
             autoplayDefault = util.toNumber( data.autoplay, 0 );
         }
 
+        if ( data.userSkipAutoplay ) {
```
henrikingo: Naming convention: You must call the attribute `data-autoplay-something=""`

See https://github.com/impress/impress.js/blob/master/src/plugins/README.md#structure-naming-and-policy

src/plugins/autoplay/autoplay.js
```diff
@@ -88,7 +110,18 @@
         }
 
         if ( timeout > 0 ) {
-            timeoutHandle = setTimeout( function() { api.next(); }, timeout * 1000 );
+            timeoutHandle = setTimeout( function() {
+                if (status === "delayed") {
+                    if (userSkipAutoplay > 0) {
+                        userSkipAutoplay = userSkipAutoplay - 1;
+                        setAutoplayTimeout( timeout );
```
henrikingo: I don't understand this... If I have

    data-autoplay="5" data-autoplay-userdelay="2"

...then the total time on the slide is set to 15 seconds? This is a very weird outcome.

A simpler alternative, both for users and as implementation would be to just have userdelay as a boolean:

    data-autoplay="5" data-autoplay-userdelay="1"    (or "true")

and then in `userActivity()` just reset the clock:

    var userActivity = function(event) {
        reloadTimeout(event);
    }

You could make this enabled by default. It seems like a good feature.

Issue #690, opened 2018-04-08
Impress.js odd touch/swipe in iOS Safari browser, fallback mode not initiating


ahainen:
Ahoy - I'm trying this on an iPhone 6, iOS v11.2.6 15D100. I noticed as well that https://impress.js.org doesn't work in iOS Safari. i.e. It's not popping fallback mode and swiping makes the images scale/slide bizarrely. 

I'm fine getting impress.js to Fallback mode on mobile, but I can't seem to do that at all :/ https://ahainen.com/akroma/roadmap-2018-q2

Example: https://i.imgur.com/bMNcvvt.gifv

Let me know if I can provide more details

henrikingo:
Also I have seen on some of my devices that the steps zoom to something really big during a swipe. Once the transition completes, the end result is correctly sized though. Is this the symptom you're reporting too?

Btw, to force the fallback mode, you can nowadays "unload" impress with such  a link:

    <a href="#" onclick="impress().tear();">Fallback mode / Disable 3D</a>

tonystar:
> Also I have seen on some of my devices that the steps zoom to something really big during a swipe. Once the transition completes, the end result is correctly sized though. Is this the symptom you're reporting too?
Yes, I think this is what was meant by @ahainen. Same to me.

PR #736, opened 2019-06-11
Bugfix: Scaling bug on swipe, when window scale is not 1

Lacuno:
When swiping to the next page on mobile, the scaling on the current page is massive and interpolates to the correct scale of the next page.
In the code the current page's scale is not corrected for the window's scale, which is corrected in this MR.

For reproducability, visit the demo (https://impress.js.org/) on mobile and swipe to the next page.
Issue that relates to the problem:  https://github.com/impress/impress.js/issues/690


henrikingo:
Very cool! Thanks for this fix.

PR #748, opened 2019-12-07
Fix typo in doc: autostart -> autoplay

tobiasBora:
The documentation of the media plugin uses `data-media-autostart` instead of `data-media-autoplay`.


henrikingo:
Thanks!

PR #753, opened 2020-02-16
Add "touchY" plugin

ArseniiLyzenko:
Added a plugin that supports up/down swipes and touches to navigate. This is just a modification of the "touch" plugin.

### DIFF
--- a/build.js
+++ b/build.js
@@ -27,6 +27,7 @@ files.push('src/plugins/autoplay/autoplay.js',
            'src/plugins/stop/stop.js',
            'src/plugins/substep/substep.js',
            'src/plugins/touch/touch.js',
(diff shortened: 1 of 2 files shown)


henrikingo:
Hi Arsenii and thanks for contributing

It seems this is merely copying the functionality of the touch plugin. That doesn't seem like something I want to merge. Am I missing something?

henrik

ArseniiLyzenko:
henrik, you kinda right, this plugin adds support vertical swipes, and it is the slight modification of "touch" plugin that supports only horizontal swipes. So, the functionality of the original plugin "touch" and "touchY" the same, except directions of swipes.

henrikingo:
Why wouldn't you just add a configuration option to the existing plugin?

PR #763, opened 2020-03-26
Add markdown dialect attribute.

### LINKED ISSUE #724: Allow Maruku dialect for Markdown
It would be nice to be able to use the Maruku dialect from the markdown-js library instead of Gruber this provides some extra functionality such as tables and even some meta handling that would potentially allow the power to have presenter notes within markdown

mrf345:
Adding a markdown dialect attribute and passing it to `markdown-js` . Resolves #724

Thanks for the awesome framework!

### DIFF
--- a/examples/markdown/index.html
+++ b/examples/markdown/index.html
@@ -120,6 +120,11 @@ are converted to HTML first, then colored by Highlight.js:
   the content was written in simple Markdown.
 </div>
 
+<div class="step slide title markdown" data-rotate="-85" data-x="4000" data-y="2000" data-markdown-dialect="Maruku">
+This is an example of a slide with a specific Markdown dialect.
+===============================================================
+</div>
+
 <script type="text/javascript">
 var enableBwCss = function(){
     disableDevopsCss();
--- a/js/impress.js
+++ b/js/impress.js
@@ -1566,10 +1566,11 @@
             var markdownDivs = document.querySelectorAll( ".markdown" );
             for ( var idx = 0; idx < markdownDivs.length; idx++ ) {
               var element = markdownDivs[ idx ];
+              var dialect = element.dataset.markdownDialect;
 
               var slides = element.textContent.split( /^-----$/m );
               var i = slides.length - 1;
-              element.innerHTML = markdown.toHTML( slides[ i ] );
+              element.innerHTML = markdown.toHTML( slides[ i ], dialect );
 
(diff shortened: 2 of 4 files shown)


js/impress.js
```diff
@@ -1566,6 +1566,7 @@
             var markdownDivs = document.querySelectorAll( ".markdown" );
             for ( var idx = 0; idx < markdownDivs.length; idx++ ) {
               var element = markdownDivs[ idx ];
+              var dialect = element.dataset.extrasMarkdownDialect;
```
henrikingo: Hi. You need to make this change in `src/impress.js` and then run `npm run build`.

examples/markdown/index.html
```diff
@@ -120,6 +120,11 @@
   the content was written in simple Markdown.
 </div>
 
+<div class="step slide title markdown" data-rotate="-85" data-x="4000" data-y="2000" data-extras-markdown-dialect="Maruku">
+This is an example of a slide with a specific Markdown dialect.
```
henrikingo: Hmm...

You have done the correct thing, but now I'm thinking maybe we should just call the attribute `data-markdown-dialect`.

examples/markdown/index.html
```diff
@@ -120,6 +120,11 @@
...
+===============================================================
+</div>
+
```
henrikingo: Can you also document the new attribute in src/plugins/README.md

examples/markdown/index.html
mrf345: i like the second one more, just did.

js/impress.js
```diff
@@ -2807,22 +2808,22 @@
  *    overwritten for individual steps and media.
  *
  *    Examples:
- *    - data-media-autostart="true" data-media-autostop="true": start media on enter, stop on
+ *    - data-media-autoplay="true" data-media-autostop="true": start media on enter, stop on
```
mrf345: `autostart` was changed automatically to `autoplay` after `npm run build`. I'm doing something wrong perhaps 🤔

henrikingo: No, it's not you. A previous commit forgot to run build.js. It is fixed in HEAD now: https://github.com/impress/impress.js/commit/99484d11b855ebd8d5f726023ca79b88e5c42d18

You can ignore this, I think it should go away in merge. Alternatively, you can do a git merge from master to get rid of it.

PR #767, opened 2020-04-15
Added the plugin for substep effects

tonegas:
This plugin allows to create easy animation using substeps.
You can show, hide or applied custom css to other objects at every substep. 
The plugin has a presentation that show all these functionalities.

### DIFF
--- a/src/plugins/subeffects/README.md
+++ b/src/plugins/subeffects/README.md
@@ -1,0 +1,41 @@
+Substep Effects Plugin
+===============
+
+This plugin allow easy creation of animation bind to substeps
+
+### USAGE
+
+This plugin will do the following things:
+
+- The plugin adds the effects for the substeps. 
+        
+When an object with class substep uses one of the following attributes, the value of these attribute indicate the objects that are subjected to some effectes:
+        
+        - data-show-only = "CLASS" : The objects with class="CLASS" are shown only in the corresponding substep.
+        
+        - data-hide-only = "CLASS" : The objects with class="CLASS" are hidden only in the corresponding substep.
+        
+        - data-show-from = "CLASS" : The objects with class="CLASS" are shown from the corresponding substep until the end or "data-show-to".
+        - data-show-to = "CLASS" : It is used with "data-show-from", the objects with class="CLASS" are shown from the substep with "data-show-from" to the corresponding substep.
+
+        - data-hide-from = "CLASS" : The objects with class="CLASS" are hidden from the corresponding substep until the end or "data-hide-to".
+        - data-hide-to = "CLASS" : It is used with "data-hide-from", the objects with class="CLASS" are hidden from the substep with "data-hide-from" to the corresponding substep. 
+
+When an object with class substep uses one of the following attributes, the value of these attributes indicate the css style to apply to a certain class. In particular:
+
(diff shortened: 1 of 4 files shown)


henrikingo:
Hi tonegas

This looks like some nice effects. But unfortunately this goes against impress.js design style. If you look at other plugins, you will notice we never set element.style from JavaScript. (Pretty much the only exception is the navigation between .step's. What we want to do in impress.js for plugins like this is just to provide CSS classes or similar, that a user can then choose to apply any CSS onto - if they want to.

Here's an incomplete example of what I mean:

    diff --git a/src/plugins/substep/substep.js b/src/plugins/substep/substep.js
    index 01d5fd5..99c2191 100644
    --- a/src/plugins/substep/substep.js
    +++ b/src/plugins/substep/substep.js
    @@ -18,6 +18,7 @@
        };
    
        var activeStep = null;
    +    var previousStepClass = null;
        document.addEventListener( "impress:stepenter", function( event ) {
            activeStep = event.target;
        }, false );
    @@ -75,6 +76,15 @@
                var el = substeps[ visible.length ];
                el.classList.add( "substep-visible" );
                el.classList.add( "substep-active" );
    +            // For the current .step, add a class that indicates the id of the currently revealed .substep element.
    +            // This is similar to how core sets a class .impress-on-step1 on body element and can be used to set CSS on unrelated elements.
    +            if ( previousStepClass ) {
    +                activeStep.classList.remove( previousStepClass );
    +            }
    +            if ( el.id ) {
    +                activeStep.classList.add( "substep-on-" + el.id );
    +                previousStepClass = el.id;
    +            }
                return el;
            }
        };

You are welcome to complete this diff to a fully working implementation if you still want to.

To be clear, it's also not wrong to do what you have done in this PR. But I would consider that custom JavaScript that you use to enhance your own presentations. (Just like everyone does their own CSS, nothing wrong with that either.) But it's not something I can merge into impress.js.

tonegas:
Thanks to your reply!
First of all, I thank you for the nice javascript framework!

Ok I got your point.
If I change the plugin applying classes instead that modifing the css style will be compliant with your implementation of impress?

Because, when I started to use impress I got the feeling that many functionalities as the one that after I proposed as plugin was very useful.

What do you think about the attribute "data-style-from-CLASS" and "data-style-to-CLASS" to apply custom style css to a relative objects from-to certain substeps?

henrikingo:
> If I change the plugin applying classes instead that modifing the css style will be compliant with your implementation of impress?

yes, if you complete the example code I did above, I will merge it. I think it makes sense to add, it's a good idea.

> What do you think about the attribute "data-style-from-CLASS" and "data-style-to-CLASS" to apply custom style css to a relative objects from-to certain substeps?

You should be able to do everything with just `.substep-on-id`. You are welcome to prove me wrong.

tonegas:
I'm sorry but I did not find anything if I look for `substep-on-ID`. 
Where is the library you are referring to?

henrikingo:
I was just referring to the example diff above. Sorry if I was unclear.

tonegas:
Ok I got your point. Thanks for the help!
I'm changing the plugin in a way will be compliant.
What do you think about the possibility to provide the plugin of some css classes for some standard effects?
And is it possible to supply the plugin of a html as the one I provided to show how to use the plugin?

henrikingo:
It's good to either provide some documentation or use the plugin in a demo, probably examples/classic-slides. We can discuss the details when you send a new PR.

PR #768, opened 2020-04-27
Subeffects classes

tonegas:
I modified the plugin now is compliant with your guidelines, I hope.
The attributes that I added permit to append custom classes following the substeps action.
I think can be useful.

### DIFF
--- a/src/plugins/subeffects/README.md
+++ b/src/plugins/subeffects/README.md
@@ -9,14 +9,15 @@ The plugin adds the posibility to add custom classes at each substep to objects:
 When an object with class substep uses one of the following attributes, the following classes are added to certain objects.
 
 The attributes are:
-        - `data-addonly-TOCLASS = "CLASS"` : This attribute is used to apply certain classes in particular `CLASS`, `CLASS-base`, `CLASS-before`, `CLASS-after`, following the rules explained in section classes.
-        - `data-addfrom-TOCLASS = "CLASS"` and `data-addto-TOCLASS = "CLASS"`: this two attributes are used to apply certain classes from the substep with the attribute `data-addfrom-TOCLASS = "CLASS"` to the substep with the attribute `data-addto-TOCLASS = "CLASS"` with the rules explained in the following.
+
+- `data-addonly-TOCLASS = "CLASS"` : This attribute is used to apply certain classes in particular `CLASS`, `CLASS-base`, `CLASS-before`, `CLASS-after`, following the rules explained in section classes.
+- `data-addfrom-TOCLASS = "CLASS"` and `data-addto-TOCLASS = "CLASS"`: this two attributes are used to apply certain classes from the substep with the attribute `data-addfrom-TOCLASS = "CLASS"` to the substep with the attribute `data-addto-TOCLASS = "CLASS"` with the rules explained in the following.
 
 The classes are:
-        - `CLASS-base`: This class is added at step enter to each object reffered by `TOCLASS`
-        - `CLASS-before`: This class is added at step enter to each obbject reffered by `TOCLASS` ad removed when the substep with the attribute `data-addonly-TOCLASS = "CLASS"` or `data-addfrom-TOCLASS = "CLASS"` is reached.
-        - `CLASS-after`: This class is added after the substep with the attribute `data-addonly-TOCLASS = "CLASS"` or `data-addto-TOCLASS = "CLASS"` is reached.
-        - `CLASS`: This class is added when the substep with the attribute `data-addonly-TOCLASS = "CLASS"` is reached and removed efterwards or between the substep with the attribute `data-addfrom-TOCLASS = "CLASS"` and `data-addto-TOCLASS = "CLASS"`.
+- `CLASS-base`: This class is added at step enter to each object reffered by `TOCLASS`
+- `CLASS-before`: This class is added at step enter to each obbject reffered by `TOCLASS` ad removed when the substep with the attribute `data-addonly-TOCLASS = "CLASS"` or `data-addfrom-TOCLASS = "CLASS"` is reached.
+- `CLASS-after`: This class is added after the substep with the attribute `data-addonly-TOCLASS = "CLASS"` or `data-addto-TOCLASS = "CLASS"` is reached.
+- `CLASS`: This class is added when the substep with the attribute `data-addonly-TOCLASS = "CLASS"` is reached and removed efterwards or between the substep with the attribute `data-addfrom-TOCLASS = "CLASS"` and `data-addto-TOCLASS = "CLASS"`.
 
 ### EXAMPLE
 
--- a/src/plugins/subeffects/README.md
+++ b/src/plugins/subeffects/README.md
@@ -1,0 +1,41 @@
+Substep Effects Plugin
+===============
+
+This plugin allow easy creation of animation bind to substeps
+
+### USAGE
+
+This plugin will do the following things:
+
+- The plugin adds the effects for the substeps. 
+        
+When an object with class substep uses one of the following attributes, the value of these attribute indicate the objects that are subjected to some effectes:
+        
+        - data-show-only = "CLASS" : The objects with class="CLASS" are shown only in the corresponding substep.
+        
+        - data-hide-only = "CLASS" : The objects with class="CLASS" are hidden only in the corresponding substep.
+        
+        - data-show-from = "CLASS" : The objects with class="CLASS" are shown from the corresponding substep until the end or "data-show-to".
+        - data-show-to = "CLASS" : It is used with "data-show-from", the objects with class="CLASS" are shown from the substep with "data-show-from" to the corresponding substep.
+
+        - data-hide-from = "CLASS" : The objects with class="CLASS" are hidden from the corresponding substep until the end or "data-hide-to".
+        - data-hide-to = "CLASS" : It is used with "data-hide-from", the objects with class="CLASS" are hidden from the substep with "data-hide-from" to the corresponding substep. 
+
+When an object with class substep uses one of the following attributes, the value of these attributes indicate the css style to apply to a certain class. In particular:
+
+        - data-style-only-CLASS = "STYLE_LIST" : Apply to objects with class=CLASS the css style="STYL_LIST" only in the corresponding substep.
+
+        - data-style-from-CLASS = "STYLE_LIST" : Apply to objects with class=CLASS the css style="STYL_LIST" from the corresponding substep until the end or "data-syle-to-CLASS".
+        - data-style-to-CLASS = "STYLE_LIST" : It is used with "data-syle-from-CLASS", the objects with class="CLASS" are setted to style="" or if "data-style-base='LIST_STYLE_BASE'" is configured in the object with class=CLASS the style is setted to style="LIST_STYLE_BASE".
+
+### EXAMPLE
+
+Some examples of all these features are presented in the file example.html
+
+### AUTHOR
+
+Copyright 2020 Gastone Pietro Rosati Papini (@tonegas)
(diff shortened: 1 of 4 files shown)


tonegas:
The example is not finished yet, because there is some text to be changed but the functionalities work.

henrikingo:
Hi.

This looks pretty much like your previous submission still. I tried to read it, but I can't find the example I suggested at all. Can you point me to it?

tonegas:
In this version of the plugin I no longer modify the css of the objects but I provide the possibility to add custom classes to the desired objects as I explain in the documentation.

https://github.com/tonegas/impress.js/blob/subeffects-classes/src/plugins/subeffects/README.md

I still have to update the text inside the file example.html, but the file uses the new paradigm.
if the plugin will be to your liking I will also change the example correctly.

henrikingo:
> In this version of the plugin I no longer modify the css of the objects but I provide the possibility to add custom classes to the desired objects as I explain in the documentation.

Yes I can see that. But other than that, it's still the same level of complexity. My suggestion was to just add a single class substeb-on-ID (where ID is the id attribute of the current substep).

tonegas:
I think your solution it easy, and clean, but less powerful.
These are some drawbacks:
1. I think it is not possible apply effect to object that are out of the current step, (possible solution could be to apply to object impress the class substeb-on-ID);
2. you need to have in your css many different rules for each substep that you use to apply effect, and I think is more complex that work in the html only;
3. you can't have a simple css rule that affect many objects with the same effect.
I think it is the main problem;
5. you need to choose different name of each sub-step;
6. I think it is complex to understand what happen you need to read css and html (in my solution only html);
7. finally with my implementation you can also apply effect before certain sub-step and after that without adding any code.

The idea of my implementation is: after you created your css effects, you can apply them on the html without looking the css. 

I try to compare the two approaches without any prejudiced:
In my approach the css classes are reusable, but in my solution the html is a bit more complex, but I think clear, see:

```
<ul>
    <li class="substep" 
    data-addfrom-obj0="substep-show-element" 
    data-addonly-obj1="substep-color-yellow">
        Step 1: Show all element and Obj1 turns yellow and Obj2 get white border
    </li>
    <li class="substep" 
    data-addfrom-obj1="substep-color-red" 
    data-addonly-obj2="substep-border">
        Step 2: Obj1 turns red from yellow and Obj2 border turns yellow and thick
    </li>
    <li class="substep" 
    data-addonly-obj3="substep-color-red">
        Step 3: Obj1 still red, Obj3 turns red and Obj2 border turns thin
    </li>
</ul>
<ul>
    <li class="obj0 obj1">Obj1</li>
    <li class="obj0 obj2">Obj2</li>
    <li class="obj0 obj3">Obj3</li>
</ul>

.substep-show-element-base{
    transition: opacity 1s ease 0s, color 1s ease 0s; 
    opacity: 0;
}

.substep-show-element{
    opacity: 1;
}

.substep-color-yellow{
    color: yellow;
}

.substep-color-red{
    color: red;
}

.substep-border-before{
    border: 1px solid white;
}

.substep-border{
    border: 2px solid yellow;
}

.substep-border-after{
    border: 1px solid yellow;
}
```
Your approach the css classes are only for this slide because id is unique, and you need to write always some css code (I'm right?):
```
<ul>
    <li class="substep" id="step1" >
        Step 1: Show all element and Obj1 turns yellow and Obj2 get white border
    </li>
    <li class="substep" id="step2" >
        Step 2: Obj1 turns red from yellow and Obj2 border turns yellow and thick
    </li>
    <li class="substep" id="step3" >
        Step 3: Obj1 still red, Obj3 turns red and Obj2 border turns thin
    </li>
</ul>
<ul>
    <li class="obj0 obj1">Obj1</li>
    <li class="obj0 obj2">Obj2</li>
    <li class="obj0 obj3">Obj3</li>
</ul>

.obj0{
    transition: opacity 1s ease 0s, color 1s ease 0s; 
    opacity: 0;
}

.step1 .obj0,
.step2 .obj0,
.step3 .obj0 {
    opacity: 1;
}

.step1 .obj1{
    color: yellow;    
}

.step2 .obj1,
.step3 .obj1,
.step3 .obj3 {
    color: red;    
}

.step1 .obj2 {
    border: 1px solid white;
}

.step2 .obj2 {
    border: 2px solid white;   
}

.step3 .obj2 {
    border: 1px solid yellow;  
}

```

Please tell me if I understand well your solution, and if the comparison is clear.

henrikingo:
>    I think it is not possible apply effect to object that are out of the current step, (possible solution could be to apply to object impress the class substeb-on-ID);

That is correct. I think the correct scope for anything related to substeps is inside the same step. Technically it would be possible to attach the class to `#impress` or `body` object, but I think that would be wrong. If you need to do similar CSS effects outside the step, then use the existing body.impress-on-ID (where ID is a step id).


>    you need to have in your css many different rules for each substep that you use to apply effect, and I think is more complex that work in the html only;

This depends. For some simple thing, like show one additional div for one substep, also CSS is simple. For a complex thing, CSS is complex. I prefer this over your PR, where your code is always complex.

I understand what you are saying here. The same limitation exists for the current `body.impress-on-ID`. We could add some improvement here, but I don't think this PR is the right solution. For steps, an idea that has been discussed before is to add functionality to group several steps into sections. Then the same mechanism could be extended to target CSS for a specific group/section too. For example: `body.impress-on-SECTION`.

>    you can't have a simple css rule that affect many objects with the same effect.
>   I think it is the main problem;

    #stepid.substep-on-onesubstep many-objects {background-color: red;}

    <ul>
    <li id="onesubstep">one substep</li>
    </ul>
    <div class="many-objects">foo</div>
    <div class="many-objects">bar</div>

>    you need to choose different name of each sub-step;

Yes. That is how id's work.

>    I think it is complex to understand what happen you need to read css and html (in my solution only html);

In your PR your html is full of custom tags, even after seeing this PR twice I can't remember what the different data- attributes mean. And your JavaScript is complex. This is not elegant code that can be merged into impress.js, I'm sorry.

>    finally with my implementation you can also apply effect before certain sub-step and after that without adding any code.

This is a special case of the second bullet point.

tonegas:
> > I think it is not possible apply effect to object that are out of the current step, (possible solution could be to apply to object impress the class substeb-on-ID);
> 
> That is correct. I think the correct scope for anything related to substeps is inside the same step. Technically it would be possible to attach the class to `#impress` or `body` object, but I think that would be wrong. If you need to do similar CSS effects outside the step, then use the existing body.impress-on-ID (where ID is a step id).

I agree.

> > you need to have in your css many different rules for each substep that you use to apply effect, and I think is more complex that work in the html only;
> 
> This depends. For some simple thing, like show one additional div for one substep, also CSS is simple. For a complex thing, CSS is complex. I prefer this over your PR, where your code is always complex.

Because I like impress.js because I can do very nice and animated slides if I do just simple things I can use latex instead. So I like to have something that avoid as much as possible to writing code when I code my presentation.

> 
> I understand what you are saying here. The same limitation exists for the current `body.impress-on-ID`. We could add some improvement here, but I don't think this PR is the right solution. For steps, an idea that has been discussed before is to add functionality to group several steps into sections. Then the same mechanism could be extended to target CSS for a specific group/section too. For example: `body.impress-on-SECTION`.

Ok, this I think go in the right way.

> 
> > you can't have a simple css rule that affect many objects with the same effect.
> > I think it is the main problem;
> 
> ```
> #stepid.substep-on-onesubstep .many-objects {background-color: red;}
> 
> <ul>
> <li id="onesubstep">one substep</li>
> </ul>
> <div class="many-objects">foo</div>
> <div class="many-objects">bar</div>
> ```
> 

Your example is not complete because you are just modifying two objects at the same sub-step. 
If you want to have the same effect on another slide or substep you need to edit the css adding a new rule.

> > you need to choose different name of each sub-step;
> 
> Yes. That is how id's work.

If I have thousands of sub-steps in my presentation I don't want create thousands of rules in my css, It is very annoying, isn't it? Probably using sections in some way you can decrease the number of rules but each new section needs a new rule.

> 
> > I think it is complex to understand what happen you need to read css and html (in my solution only html);
> 
> In your PR your html is full of custom data- attributes. Even after seeing this PR twice I can't remember what the different data- attributes mean. And your JavaScript is complex. This is not elegant code that can be merged into impress.js, I'm sorry.
> 

Ok on this point. I agree, I'm not used to write js so, you are right. But I don't like also your solution because I need to write many lines of css. I can't have standard effect that simply apply on my objects without writing css. 

> > finally with my implementation you can also apply effect before certain sub-step and after that without adding any code.
> 
> This is a special case of the second bullet point.

You are right.

The point is I would like to have something that without any additional css  writing (e.g. I download a standard effects library) permits to add these effects on my presentation, following the sub-steps rules. I agree that my code is not clean us yours but your solution it is not what was I looking for. 

Can we modify your solution to get closer to my idea? 
Do you think we can find a middle way?

henrikingo:
> Because I like impress.js because I can do very nice and animated slides if I do just simple things I can use latex instead. So I like to have something that avoid as much as possible to writing code when I code my presentation.

This is indeed a benefit of impress.js. To be clear, it's great that you are using this javascript to add effects to your presentations. That's exactly the idea. But to be added to impress.js itself, a certain generalism and simplicity is required. I think your proposal is very specific to something you wanted to do.

> The point is I would like to have something that without any additional css writing (e.g. I download a standard effects library) permits to add these effects on my presentation, following the sub-steps rules. 

I'm afraid I have bad news: the design goal in impress.js has always been that impress.js provides the javascript code and the author provides all the CSS. (Even if you probably copy a lot of the demo css, technically none of the provided css is supposed or required to be used.)

It is ok for you to mix additional js libraries, but impress.js will not add code to support that. Otoh, as your code shows, you can use the events from impress.js to integrate and extend it. (Similar to css classes, we can add more events if there is something you need.)

> If I have thousands of sub-steps in my presentation I don't want create thousands of rules in my css, It is very annoying, isn't it? Probably using sections in some way you can decrease the number of rules but each new section needs a new rule.

Hopefully the sections would somehow correlate with how you use the effects, so that the same section would use a given set of effects and it would be natural to have different CSS rules for them.

tonegas:
> > Because I like impress.js because I can do very nice and animated slides if I do just simple things I can use latex instead. So I like to have something that avoid as much as possible to writing code when I code my presentation.
> 
> This is indeed a benefit of impress.js. To be clear, it's great that you are using this javascript to add effects to your presentations. That's exactly the idea. But to be added to impress.js itself, a certain generalism and simplicity is required. I think your proposal is very specific to something you wanted to do.
> 
> > The point is I would like to have something that without any additional css writing (e.g. I download a standard effects library) permits to add these effects on my presentation, following the sub-steps rules.
> 
> I'm afraid I have bad news: the design goal in impress.js has always been that impress.js provides the javascript code and the author provides all the CSS. (Even if you probably copy a lot of the demo css, technically none of the provided css is supposed or required to be used.)
> 
> It is ok for you to mix additional js libraries, but impress.js will not add code to support that. Otoh, as your code shows, you can use the events from impress.js to integrate and extend it. (Similar to css classes, we can add more events if there is something you need.)
> 
> > If I have thousands of sub-steps in my presentation I don't want create thousands of rules in my css, It is very annoying, isn't it? Probably using sections in some way you can decrease the number of rules but each new section needs a new rule.
> 
> Hopefully the sections would somehow correlate with how you use the effects, so that the same section would use a given set of effects and it would be natural to have different CSS rules for them.

Ok, I'm very sad that you are not interested to add such functionalities. I worked to realized a framework that was compliant with your idea. I changed my first implementation moving from apply css to apply classes instead, but it seems to be not enough. 

I understand that my code is not good, I know but when I started to use impress I spent one mouth to setup. I would like just to save a bit of time to people want to use.

Thanks to your time.

PR #774, opened 2020-10-10
Support Chinese in impressConsole

kdxcxs:


henrikingo:
Why not all three?

    case zh:
    case zh-cn:
    case zh-CN:
        ...

kdxcxs:
Cause it's just simplified Chinese,considering the existence of traditional Chinese,I think it's better like that.

henrikingo:
Thanks. This is a cool addition of a major language!

kdxcxs:
My pleasure

PR #777, opened 2021-01-02
@bartaz removes himself from tutorial text ;)

bartaz:
It's been almost 9 years since when [impress.js was created](https://github.com/impress/impress.js/commit/0c7f9cc2b14f9a5c68f7847f13a7708934fe16e5). Yes, it's gonna be 10 years in December this year!

Today during some cleaning I even found some of my initial notes for impress.js and presentation!

IMG_20210102_121840

I've received lots of nice messages over these years on Twitter, but it's also quite a while since I've done anything to the project so it is right about time to remove myself from `index.html` and for sure I'm not the only one to receive the credit for it. 

Thanks everyone who contributed to the project so far!
You have less than a year to think of some 10 year birthday celebrations :D

### DIFF
--- a/index.html
+++ b/index.html
@@ -458,13 +458,10 @@ if ("ontouchstart" in document.documentElement) {
     
     Are you still reading this?
     
-    For real?
+    For real? I'm impressed! 
     
-    I'm impressed! Feel free to let me know that you got that far (I'm @bartaz on Twitter), 'cause I'd like
-    to congratulate you personally :)
-    
-    But you don't have to do it now. Take my advice and take some time off. Make yourself a cup of coffee, tea,
-    or anything you like to drink. And raise a glass for me ;)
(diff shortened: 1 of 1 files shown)


henrikingo:
Thanks :-)

The CircleCI build fails this https://github.com/impress/impress.js/blob/d0c427f172453b6306fb88d638df7d1594c476c8/.circleci/config.yml#L9-L11

I see now that the same has failed a merge 2 months ago but worked until 3 months ago. Must be something with CircleCI and using sudo. I'm not sure that step is needed at all. I can have a look at it later.

bartaz:
Thanks, I wasn't sure why would Circle CI fail on text change in HTML comment.
And I guess you shouldn't need to manually install npm on Circle CI these days.

henrikingo:
Ok can you rebase an try again.

Btw, thanks for sharing the photo!

PR #779, opened 2021-01-14
Add ability to specify substep order

codesections:
Thanks for the very helpful project!  For the presentation I'm currently writing, I realized it would be helpful to display substeps in a different order than the one in which they appear in the HTML.  (I could have hacked around that by re-positioning items with CSS, but that would be pretty ugly and not at all semantic).

To fix my issue, I added a feature to the Substep plugin and decided to send it upstream as a PR in case others would also find it useful.

This PR allows users to specify a custom order via the `data-substep-order` attribute.  Substeps without a
`data-substep-order` attribute are revealed last.

This PR also updates the Substep README to document the new feature.

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -2081,6 +2081,20 @@
             'useAMPM': false
         };
         break;
+    case 'zh-CN':
+    case 'zh-cn':
+        lang = {
+            'noNotes': '<div class="noNotes">当前帧没有备注</div>',
+            'restart': '重新开始',
+            'clickToOpen': '点击以打开演讲者控制界面',
+            'prev': '上一帧',
+            'next': '下一帧',
+            'loading': '加载中',
+            'ready': '就绪',
+            'moving': '移动中',
+            'useAMPM': false
+        };
+        break;
     case 'en': // jshint ignore:line
     default : // jshint ignore:line
         lang = {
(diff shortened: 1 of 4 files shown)


henrikingo:
Thanks.

On first glance this looks like a well done and reasonable patch.

It was indeed the intention that you should define the order in HTML and then use CSS to move around elements if you don't want them to appear top down. I'll think about your proposal for a while though...

To help convince me, would you like to share a link to the presentation where you have used this feature?

codesections:
I'd be happy to.  The presentation is [Imagining the Ideal Language for Writing Free Software](https://codesections.com/presentations/fosdem-2021/), which I'll be giving tomorrow in the [Perl and Raku devroom](https://fosdem.org/2021/schedule/track/perl_and_raku_programming/) at FOSDEM.  A good example of the usecase I have in mind shows up in [exciting languages](https://codesections.com/presentations/fosdem-2021/#/exciting-languages) slide, where I go through three bullet points and then revisit each bullet to add aditional info.  The HTML for that list (with some styling classes removed) currently looks like this:

```html

 <ul>
     <li class="substep"  data-substep-order="0">
         TypeScript <span class="substep" data-substep-order="3">(Microsoft)</span>
     </li>
     <li class="substep"  data-substep-order="1">
         Golang <span class="substep" data-substep-order="4">(Google)</span>
     </li>
     <li class="substep data-substep-order="2">
         Rust<span class="substep" data-substep-order="5">(Mozilla et al.)</span>
     </li>
 </ul>
```

I _could_ achieve something visually similar with CSS, of course but (imo) the patch in this PR makes the semantics of the slide more clear: it really is just one list of bullets, after all.

I use the feature again [two slides later](https://codesections.com/presentations/fosdem-2021/#/big-tech-vs-free-software) in the inverse way: there I have two bulleted lists where I want to display item 1 from the left list, then the matching item from the right list and so on for each pair.  Again, this would be possible to achieve via CSS, but I'd lose the semantic info of these being two lists (which is clear after all steps are displayed).

henrikingo:
Thanks. That looks nice and I'm glad to hear you used impress.js at Fosdem! Thanks for this contribution.

PR #781, opened 2021-02-20
Add roam plugin.

kdxcxs:
Add roam plugin to allow the user to change the view freely.


src/plugins/roam/roam.js
```diff
@@ -0,0 +1,254 @@
...
+            start: undefined
+        };
+    } );
+
+    // Request to be called before next frame renders.
+    // And every single frame after is going to be called in the `roamAnimationFrame()` function.
+    window.requestAnimationFrame( roamAnimationFrame );
+
+    // Make the keys controlling.
+    document.addEventListener( "keydown", eventHandler );
```
henrikingo, line 252: You should add event listeners via the gc helper. This will keep a reference to them and remove them when `impress().tear()` is called.

https://github.com/impress/impress.js/blob/master/src/lib/gc.js

For example:
https://github.com/impress/impress.js/blob/4c9d2e38aaf1ea038118d4ff6549609df6f1fab3/src/plugins/blackout/blackout.js#L93

You also need to do these initializations from an event listener listening to `impress:init`.

src/plugins/roam/roam.js
```diff
@@ -0,0 +1,254 @@
...
+    roams.keys = "";
+    Array.from( roamKeys ).forEach( ( key ) => {
+        roams[ key ] = {
+            start: undefined
+        };
+    } );
+
+    // Request to be called before next frame renders.
+    // And every single frame after is going to be called in the `roamAnimationFrame()` function.
+    window.requestAnimationFrame( roamAnimationFrame );
```
henrikingo, line 249: See comment below. I don't have a specific suggestion, but also here there should be a mechanism where this plugin becomes passive when `impress().tear()` is called.

src/plugins/roam/roam.js
```diff
@@ -0,0 +1,254 @@
...
+            }
+        } );
+
+        // Everything's done.
+        // Just leave the impress to update the "canvas" and we've finished the work for one frame!
+        // As there no more frame between present one and the comming one, duration should be 0.
+        api.goto( presentStep.id, 0 );
+    }
+
+    function eventHandler( event ) {
```
henrikingo, line 230: Please make sure this plugin allows typing text into form elements. See https://github.com/impress/impress.js/blob/master/src/plugins/form/form.js

henrikingo (changes requested):
Thanks kdxcxs!

This is very impressive. I have implemented something similar in [impressionist](https://github.com/henrikingo/impressionist/blob/master/src/plugins/camera/camera.js) however, unlike you I *did* miss CS computer graphics classes and you can tell the difference. I was impressed to read your clean implementation. 

I have a small number of impress.js specific changes requested.

In addition, it could be nice to advertise this feature in some of the demo presentations. The classic-slides demo is intended to showcase most features, so you could add text somewhere - maybe to the last overview step - to advertise roam plugin. In addition it seems the 3D-rotations demo is kind of topical here?

src/plugins/roam/roam.js
kdxcxs: OK. I'll fix them. Thanks for mentioning.

kdxcxs:
Actually I'm not sure if it's better to move just the view instead of the active step.

Also, I couldn't come up with a impressive way to advertise it for now as I want to keep the demo cool and clean. I'll do my best.

henrikingo:
> Actually I'm not sure if it's better to move just the view instead of the active step.

Of course. I just meant the overview step could have the text somewhere. Or it can just be a regular slide.

> Also, I couldn't come up with a impressive way to advertise it for now as I want to keep the demo cool and clean. I'll do my best.

I can do it separately if you don't feel like it.

src/plugins/roam/roam.js
```diff
@@ -236,19 +247,35 @@
...
-    // Request to be called before next frame renders.
-    // And every single frame after is going to be called in the `roamAnimationFrame()` function.
-    window.requestAnimationFrame( roamAnimationFrame );
+        // Stop requesting animation frame when `impress().tear()` is called.
+        api.tear = new Proxy( api.tear, {
+            apply ( target, thisArg, argumentsList ) {
+                impressActive = false;
+                target.apply( thisArg, argumentsList );
+            }
+        } );
```
henrikingo, line 261: Ah, sorry. I should have pointed to this in my last review, but I had forgotten how exactly to do this. Please use

    gc.pushCallback(my_function);

https://github.com/impress/impress.js/blob/f68e5fff298469cf9836fa6121f665d8ab323e2a/src/lib/gc.js#L60

... to register a function that is called as part of `impress().tear()`.

Example:

https://github.com/impress/impress.js/blob/d0c427f172453b6306fb88d638df7d1594c476c8/src/plugins/progress/progress.js#L22

henrikingo:
Oh, and please run `npm run all` and then include `js/impress.js` in your commit. Due to backward compatibility impress.js has the built file directly in the git repo.

henrikingo:
Hi, again... So I downloaded your branch and tried it a bit. It works and is really cool, but...

I think this is what you referred to as well: It probably makes more sense not to move the current step and rather move the "view". Impress calls it the canvas element, in Impressionist I call it "camera". Anyway, this is how you get to the canvas element:
https://github.com/henrikingo/impressionist/blob/788e6d05e50287ea52911ae485a27054a46c3516/src/plugins/camera/camera.js#L82

It's a `<div>` inserted dynamically between the `div#impress` element and the `.step` elements. Moving the canvas can be done with exact same functions as you have, except I believe you have to move in the opposite direction compared to what you are doing now.

Moving the canvas means that I can "roam" around the entire presentation, but the slides don't move in relation to each other.

Minor: For the left hand, I wonder if it would be more natural to use keys a-f, w-s, e-d instead? (Corresponding to left-right, up-down, in-out, respectively.) The way you have it now, I would expect w-s to move up-down. But this is subjective, it's equally correct the way you've implemented it.)

PR #789, opened 2021-06-19
apply markdown dialect to all elements

thawk:
Dialect is only applied to the last element, so the previous elements will not has dialect.

This PR apply dialect to all markdown elements, so TABLE and other features can be used in all elements.


henrikingo:
Thanks!

PR #790, opened 2021-07-06
byId support non ascii characters to support non ascii id for steps

thawk:
``impress.js`` currently gets step id from page hash and pass it to ``byId()`` to jump to specified step.
``byId()`` uses ``document.getElementById( id )``, it works fine with ascii characters.  But for non-ascii ids, it may be url encoded, so ``byId()`` failed to get element by those ids.
This PR uses ``decodeURIComponent( id )`` to decode step id before passing to ``document.getElementById( )``, so it handles unicode step ids too.

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -1204,7 +1204,8 @@
 
             // Get id from url # by removing `#` or `#/` from the beginning,
             // so both "fallback" `#slide-id` and "enhanced" `#/slide-id` will work
-            return byId( window.location.hash.replace( /^#\/?/, "" ) );
+            var encoded = window.location.hash.replace( /^#\/?/, "" );
+            return byId( decodeURIComponent( encoded ) );
         };
 
         // `getUrlParamValue` return a given URL parameter value if it exists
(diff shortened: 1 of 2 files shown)


henrikingo:
Hi

First, note to self: Indeed HTML5 allows non-ascii characters in the id. In impress.js we target modern HTML, so this patch is admissible.

Now, I would like to change this slightly differently. IMO (yes, this is subjective, but if we follow my subjective style, at least we are consistent...) the `byId()` function should do exactly what it says and not transform the input string in any way. It is caller's responsibility to decode the URI hash first. So for example, you should call `decodeURIComponent` in `lib.util.getElementFromHash()` instead.

Unfortunately this way is a bit more complicated, because now we have to fix or check other parts that directly or indirectly call `byId()`. I grepped the src/ directory and luckily surface area is quite limited. It's mostly a case to check the use of `goto()` function in impress API.

    grep -r 'goto(' src
    grep -r 'byID' src
    grep -r getStep src

henrikingo (changes requested):
.

thawk:
You're right, we should let ``byId()`` do what it means.  I've update this PR.
The exists code work perfectly with Unicode, except for hash from URL, it works except for page reload.  So we only need to modified ``getElementFromHash()`` to make it work.

henrikingo:
Ok, that was nice and simple then :-) Thanks!

Issue #798, opened 2022-01-01
Incorrect relative width/height in the rel plugin


thawk:
In ``toNumberAdvanced()`` of ``rel.js``, it use ``window.innerWidth`` and ``window.innerHeight`` as unit for relative-to-screen-size, which is incorrect.

The number return from ``toNumberAdvanced()``, should be logic width/height.  But when the ``w``/``h`` postfix is used, the return value is calculated based on the actually screen width/height.

``1w`` should equals to one ``impress.config.width``, and ``1h`` should equals to one ``impress.config.height``, not one ``window.innerWidth``/``window.innerHeight``.

I try to fix this, but the ``rel.js`` is a pre-init plugin, when it is executed, ``impress.config`` is not initialized yet.  And the ``defaults`` of ``impress.js`` is not exported, make it impossible to calculate the screen width/height if the user doesn't specify ``data-width``/``data-height`` for the ``#impress`` element.

PR #799, opened 2022-01-02
Try fixing relative-to-screen-size calculation

thawk:
Add an unittest for #798 , and make an attampt to fix that.
Not a perfect solution, just work.

### DIFF
--- a/src/impress.js
+++ b/src/impress.js
@@ -177,6 +177,9 @@
         transitionDuration: 1000
     };
 
+    // Configuration options
+    var config = null;
+
     // It's just an empty function ... and a useless comment.
     var empty = function() { return false; };
 
@@ -227,9 +230,6 @@
 
         // Array of step elements
         var steps = null;
-
-        // Configuration options
-        var config = null;
 
         // Scale factor of the browser window
         var windowScale = null;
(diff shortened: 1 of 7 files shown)


src/impress.js
```diff
@@ -335,18 +354,9 @@
...
-                maxScale: lib.util.toNumber( rootData.maxScale, defaults.maxScale ),
-                minScale: lib.util.toNumber( rootData.minScale, defaults.minScale ),
-                perspective: lib.util.toNumber( rootData.perspective, defaults.perspective ),
-                transitionDuration: lib.util.toNumber(
-                    rootData.transitionDuration, defaults.transitionDuration
-                )
-            };
+            // Initialize configuration object.
+            // Pre-init plugins may make some change, so we calculate it again.
+            config = buildConfig();
```
henrikingo, line 359: Hmm... The way I see it, impress.getConfig() is global and any plugin can access it at any time and change the values. So this kind of arbitrarily resets it at one poínt in time, but doesn't actually protect anything in a robust way.

I see two options:

1. Make getConfig always return a new copy of the config:

    return buildConfig();

2. Just remove these lines and either expect plugins to be well behaved and not mess with the config, or alternatively consider it a feature that the config can be dynamically changed. (The latter is debatable to be fair.)

I'm ok with either so you can choose.

henrikingo (changes requested):
Thanks. Nice catch.

I have just one small change request.

src/impress.js
thawk: I choose the second option.
After some searching, I found that no plugins modifies impress config, and even no plugins other than `rel` use that.  So it's safe to assume that the config WILL NOT be changed in pre-init stage.  If there's need for modify it, we can introduce a `setConfig()`, but we should consider that invoke order to make sure the effect it config modification.  So I think is's better NOT provide the `setConfig()` by now.

henrikingo: I agree. And yes, when writing the above I already knew nobody currently modifies the config. I could have been more explicit about that.

Issue #816, opened 2022-06-26
Auto detect slide title for markdown slides


thawk:
We can use the first non-blank line for markdown title.  For a regular presentation, the first line may be suitable for title.
I think it's better than the `step-X`, and can easy navigation.

thawk:
For example, with the example `examples/markdown/index.html`, in the bottom-right `goto` dropbox, it shows `markdown`/`step-2`/`step-3`/`step-4`....  It's hard to select a correct slide to goto.

To get a meaningful title, we can use the first line of each slide.  I think the first line of a slide should be the most important sentence at most time.   With this, the titles of the above example will be `markdown`/`Markdown.js`/`Styles`/`A blockquote & image`/`Code`/`Slide separator`/`Title slide`...  I think it's much easier to goto the right slide.

-----

For the following example, there will be 3 slides, titled `Slide title`/`Sub-title`/`Another point`.

```markdown
# Slide title

Some text

-----

## Sub-title

* bullet point

## Sub-title2

* some thing

-----

Another point
```

henrikingo:
Ah, thanks for clarifying. This makes sense.

thawk:
Closed by #821

Issue #812, opened 2022-06-15
Trim common leading indention for Markdown


thawk:
Currently, additional indentions is not allowed for Markdown, it will be recognized as [indented code block](https://spec.commonmark.org/0.30/#indented-code-block).  This makes it hard to using HTML formatter to format the code.

Suggest to remove the common leading indention for a Markdown code block.  We can use the first non-blank line to detect the amount of leading spaces, than remove them before feeding to Markdown parser.

henrikingo:
Makes sense. I always write the Markdown left aligned, but a html editor may disagree with that, sure.

Oh, that reminds me... Filed #813 to capture an idea I had some years ago but apparently never created an issue.

thawk:
Closed by #817

Issue #811, opened 2022-06-15
Upgrade Markdown parser 


thawk:
It seems that `extras/markdown/markdown.js` is a quite old version.  In that file, the last updated date is marked as `2013-09-15T16:12Z`.
Suggest to upgrade it to a more modern version, or replace by another parser.
I've tried using [marked](https://github.com/markedjs/marked), it's as easy as replacing one line in `plugins/extras/extras.js`.

henrikingo:
+1
I'd welcome a PR for sure.

thawk:
Closed by #817

PR #817, opened 2022-06-26
Improve markdown support.

thawk:
1. Support `marked`, you can using it instead. #811
2. Auto detect slide title (using first non-blank line). #816
3. Auto trim the indentation, so it can be indent like regular HTML.
   #812


henrikingo:
Hmm... May I ask you to submit a separate PR for each feature / issue.

In particular, I'm not convinced about #816 but would be happy to merge the others.

thawk:
Sorry, I forgot to update #816.  I've update this PR to related to #811 and #812 only.

henrikingo:
Thanks. As far as I can see, this is ready to go as is! Nice work.

PR #821, opened 2022-08-23
Auto guess slide title from first non-blank line

thawk:
Markdown slides can be split into multiple slides by `-----`, but currently there's not way to set slide for slides other than the first one.  What this PR does is:

- If a  markdown step has title, use it for the first slide
- Otherwise use the first non-blank line as title for each splited slides

It should make slide navigation easier by a meaningful title.

### DIFF
--- a/src/plugins/extras/extras.js
+++ b/src/plugins/extras/extras.js
@@ -17,7 +17,7 @@
     const SLIDE_SEPARATOR = /^-----$/m;
 
     // Max length for title.
-    // Line longer than this won't be used as title
+    // Line longer than this will be cutted.
     const MAX_TITLE_LEN = 40;
 
     const getMarkdownParser = function( ) {
@@ -58,10 +58,9 @@
             if ( line.length > 0 ) {
                 if ( line.length <= MAX_TITLE_LEN ) {
                     return line;
+                } else {
+                    return line.slice( 0, MAX_TITLE_LEN - 3 ) + "...";
                 }
-
-                // The first non-blank is too long to be a title
-                return "";
             }
         }
         return "";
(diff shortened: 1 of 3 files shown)


src/plugins/extras/extras.js
```diff
@@ -48,6 +52,21 @@
...
+    const guessSlideTitle = function( text ) {
+        for ( var line of text.split( "\n" ) ) {
+            line = line.trim( );
+            if ( line.length > 0 ) {
+                if ( line.length <= MAX_TITLE_LEN ) {
+                    return line;
+                }
+
+                // The first non-blank is too long to be a title
+                return "";
```
henrikingo, line 64: Why not cut the line to 40 char?

henrikingo (commented):
Thanks.

This is a good idea. Especially for the navigation menu.

Just one question.

src/plugins/extras/extras.js
thawk: For a long sentence, I don't know whether it's suitable to keep the first part.  But maybe it is better than `page-n`? I'll update the PR.

henrikingo:
Great modifications, especially the last one.

Is this ready to merge? 

You know what? I would like to make you co-maintainer, and then you can merge yourself. Is that ok with you?

henrikingo (approved):

thawk:
Yes, I think it's ready to be merged.
The only problem is I can't distinct whether an id of slide like `step-N` is assign by the author or by impress.js while initializing.  We need to store the original attribute of each slide.  But I think authors may not manual assign id like `step-N`.

henrikingo:
I think that's ok :-)

If you accept the invites to the Admins and Mergers groups, you should be able to merge this PR yourself.

PR #827, opened 2022-11-22
Add .slideView or .preView class if loaded in console

fnogatz:
The impress console is a really useful tool for presenters. As of now, it is however not possible to change the slide's style only for the *slideView* (current slide) or only for the *preView* (next slide). We can't use CSS here as it is not aware of the fact if the page is loaded in an iframe.

The `impressConsole.js` already dynamically adds the class `impress-console` to the body of both iframes. This PR additionally adds the `slideView` and `preView` class, respectively. As a result, we can for instance hide the progress, progressbar and toolbar only in the *preView* using the following CSS snippet:

```css
.impress-console.preView .impress-progress,
.impress-console.preView .impress-progressbar,
.impress-console.preView #impress-toolbar {
  display: none;
}
```

A real-world example that uses the proposed change can be found in the slides' console of my [PhD defense](https://nogatz.net/talks/202210_phd-defense/) ([source code](https://github.com/fnogatz/talks/tree/gh-pages/202210_phd-defense)).

### DIFF
--- a/css/impress-common.css
+++ b/css/impress-common.css
@@ -94,6 +94,23 @@ It is focused on plugin functionality, not the visual style of your presentation
 }
 
 /*
+  We might want to hide the help, toolbar, progress and progress bar in the
+  preView window of the impressConsole that is displayed when you press P.
+*/
+.impress-console.preView .impress-progress,
+.impress-console.preView .impress-progressbar,
+.impress-console.preView #impress-toolbar,
+.impress-console.preView #impress-help {
+  display: none;
+}
+/*
+  Hide the help in the slideView as well.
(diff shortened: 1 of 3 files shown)


henrikingo:
Thanks again Falco! In fact, this has bothered myself too I think.

Would you like to add your example CSS to `css/impress-common.css`. That way users will see it - and let's face it, copy it - from there.

fnogatz:
Thank you for the feedback, @henrikingo! I added the CSS as suggested in `css/impress-common.css`, together with some little explanation.

henrikingo (approved):

PR #831, opened 2022-11-29
Getting Started Guide 

simplePCBuilding:
For me, getting started with impress.js was not that easy, honestly, in retrospective, I don't know why. So I created a little "Getting Started" guide to help new users starting out. I also noticed a problem, where the notes didn't hide as they should have. So I added some code that takes care of that (it just sets the CSS display property to "none" on all objects with class "notes". Tested it, it worked.)

I will also be creating an introduction video in a short while and link it in the Getting Started Guide.


simplePCBuilding:
I will fix the issues rn

src/plugins/impressConsole/impressConsole.js
```diff
@@ -16,6 +16,12 @@
 ( function( document, window ) {
     'use strict';
 
+    var notes = window.document.getElementsByClassName( 'notes' );
+
+    for ( let items in Object.entries( notes ).values ) {
+        items.style.display = 'none';
+    }
+
```
henrikingo, line 24: This seems to be a bug fix in its own right! Can you please file it as a separate PR and undo in this PR.

henrikingo (commented):
Hey thanks this is amazing! I will have to read it with thought tomorrow. Thanks for writing this!

simplePCBuilding:
I have not really used PRs before, so I'm not sure how to have multiple PRs from the same repo (so my fork where I pushed the changes). It appears that whenever I push to that repo, the PR updates... I will look up how to do it today and will update the PR asap

simplePCBuilding:
I just got it... it works by adding additional branches right?

henrikingo (commented):
Thank you! 

I made a few comments which you may or may not want to incorporate. Since this is not a code review, I'm fine leaving the final editorial judgement to you, but wanted to add some additional viewpoints where I had any.

GettingStarted.md
```diff
@@ -0,0 +1,133 @@
+# Introduction
+Welcome to impress.js! This presentation framework allows you to create stunning presentations with the power of CSS3 transformations. It is inspired by the idea behind [prezi.com](https://prezi.com). 
+**NOTE:** This Guide is not made for you, if you have never written HTML and/or CSS before. Knowing your way around in JavaScript certainly helps, but is not a necessity, as we won't go over how you may use JavaScript in conjunction with this framework here. You may still continue this tutorial and try to understand what we do as you go.
```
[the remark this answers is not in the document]
simplePCBuilding: right. Will add this.

GettingStarted.md
```diff
@@ -0,0 +1,133 @@
...
+
+# Getting started with impress.js
+## Installation / acquiring the framework
+You first need to decide if you want to use the bleeding edge (most up to date, but higher risk of experiencing problems) or the stable release. 
+
+### **Stable release:**
+Head to the releases tab and download the source code as zip or as a tarball. Go ahead and unzip / untar it. You may find the framework under ```/js/impress.js.```
+
+### **Bleeding Edge:**
+In the github repository, click the "Code"-Button, then click "Download ZIP". Go ahead and unzip / untar it. You may find the framework under ```/js/impress.js.```
```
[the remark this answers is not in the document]
simplePCBuilding: added.

GettingStarted.md
```diff
@@ -0,0 +1,133 @@
...
+    <script>impress().init()</script>
+</body>
+</html>
+```
+
+Now, head into a file-manager, navigate to the file you just created (*index.html*) and open it. You should end up in a browser where you should see "My first Slide" displayed. As this is not really exciting, we are not gonna change anything about that and are gonna look at what you just typed. What do these lines mean?
+
+Well, first things first, you should probably give your presentation a title. You may do this in normal HTML fashion by changing the *title* HTML tag.
+
+So not, we reached the HTML body. You can see that it already belongs to a class. This class just tells impress.js that this is the body where the "fallback-message" should be displayed when it detects, that your browser does not support CSS3 and therefore impress.js won't work. You can easily omit that class though, including the "fallback-message" div with its content, if you only intend to use the presentation for yourself and you know about the fact that some browsers might not work.
```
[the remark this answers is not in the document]
simplePCBuilding: now* fixed that one

GettingStarted.md
```diff
@@ -0,0 +1,133 @@
...
+data-rotate-x   | Rotate the element along the x-axis
+data-rotate-y   | Rotate the element along the y-axis
+data-rotate-z   | Rotate the element along the z-axis
+data-scale      | Scale the element. 
+
+These are the basic positioning options in impress.js. All of the attributes take Strings as arguments, so be aware of the fact, that you need to put quotation marks around the numbers! The *data-rotate* attributes take an angle in form of a String as argument.
+
+Now, that you have created the slides, you might want to style them. This is where CSS comes into play. Add another file to your project called, e.g., ```style.css```. 
+
+**NOTE:** It is not intended that you modify the CSS properties of the built-in class *step*. Instead add another class to that element and style that class. See the example below.
```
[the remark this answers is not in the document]
simplePCBuilding: I will rephrase it a little to make it more clear that you can but shouldn't

GettingStarted.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Goto
+This plugin allows you to directly go to a certain step, by either passing in a number or the id of the step you'd like to go to.
+
+## Progress
+This plugin, as its name implies, displays the progress in your presentation.
+
+## Blackout
+This plugin hides the screen, if you press B, which is handy in a lot of situations.
+
```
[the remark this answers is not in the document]
simplePCBuilding: done

simplePCBuilding:
I've just come up with a great idea to make the usage easier. Afaik, you can include a js file from raw.githubusercontent.com... If this is so, one can just copy that link with a script tag into the head of the html file and therefore you don't have to download it.

simplePCBuilding:
Doesn't appear to work. I might be able to pull something off that is slightly different. I have only tested it on my iPad so this might be the problem as well.

simplePCBuilding:
https://cdn.jsdelivr.net/npm/impress.js@1.1.0/js/impress.js

There we go. jsdelivr is the solution

simplePCBuilding:
This can't possibly be my code. I have only modified the README and added the Getting Started guide

henrikingo:
> I've just come up with a great idea to make the usage easier. 

...and you have just discovered what I call "Documentation Driven Development": When someone starts writing docs for a software project, but realizes "this is stupid, it would be much better/easier/elegant if it worked like this instead". Then implement the new idea, as it not only makes the software better, but also the documentation!

> Afaik, you can include a js file from raw.githubusercontent.com...

Yes, but hold your horses. The fact that you put a copy there proves that it works. But, what happens when someone does the next commit? What we need now is a deployment pipeline. Which hopefully can be implemented with github actions. In other words the online .js file should be updated automatically whenever we merge a new PR. And whenever we make a release, it should publish a versioned file representing the release. Like jsdelivr does I suppose.

Can you please file a separate ticket for that? And of course feel free to implement it

simplePCBuilding:
I can. There's actually no need to implement anything. Jsdelivr does this automatically.

henrikingo:
> What happened?

I'm not sure but 214ce132e3e70b64d1731d1602462e416ed3ab34 doesn't exist aywhere in Github. Did you perhaps rebase or force push something  where a previous commit was overwritten?

simplePCBuilding:
I have not, there are no other changes. I think it may be due to the links...

henrikingo:
Ok you're right, the commit exists in your fork. Don't know why github didn't find it when I searched.

GettingStarted.md
```diff
@@ -0,0 +1,152 @@
...
+        <p>For the best experience please use the latest <b>Chrome</b>, <b>Safari</b> or <b>Firefox</b> browser.</p>
+    </div>
+
+    <div id="impress">
+        <div id="myFirstSlide" class="step">
+            <h1>My first Slide</h1>
+        </div>
+    </div>
+
+    <script type="text/javascript" src='./impress.js></script>
```
henrikingo, line 53: Shouldn't you also use the online file (jsdelivr) here too?

GettingStarted.md
```diff
@@ -0,0 +1,152 @@
...
+- Source: https://cdn.jsdelivr.net/gh/impress/impress.js/js/impress.js
+
+For older versions, please just replace the version number behind the @!
+
+**Loading the script directly from cdn**
+Loading the script from the cdn is quite straight forward. You need to include a script tag in the head of the HTML document:
+
+```
+<script src="https://cdn.jsdelivr.net/gh/impress/impress.js@2.0.0/js/impress.js"></script>
+```
```
henrikingo, line 23: This is a great idea.

Except... An important use case for presentations is that you're traveling somewhere and presenting in front of an audience, and no internet.

So... This addition is okay and it simplifies life for many. But we should still make a statement about offline use and self contained presentations. Now... We are back to where we started: Do you write more documentation, or can we somehow simplify impress.js to keep it elegant? It seems the following is possible:

    <script src="https://cdn.jsdelivr.net/gh/impress/impress.js@2.0.0/js/impress.js"></script>
    <script>window.impress || document.write('<script src="js/impress.js">\x3C/script>')</script>

This would first try an online URL for impress.js, but if it didn't load, it fall back to trying a local copy of the same file.

henrikingo (commented):
Ok so looks good otherwise, but now we can fine tune the jsdelivr version of impress.js. This conversation can also continue in the separate ticket.

GettingStarted.md
simplePCBuilding: Yeah this is just meant for ease of use, if you have WiFi. If you don't, you can obviously just add the js file directly. I personally feel like we should add the js file in the releases tab as a direct download, as this makes the download faster

GettingStarted.md
simplePCBuilding: Technically yes. I thought, I could show both versions.

henrikingo: I think it's better to be consistent.

GettingStarted.md
henrikingo: That's a good suggestion...

simplePCBuilding:
I just changed the wording regarding jsDelivr to better explain the usage.

simplePCBuilding:
I feel like we should also start to provide .min.js files for the cdn. I could update the package.json to include minify (or browserify for that matter)

henrikingo:
Note that build.js already creates a minified file. It's just not committed to the git repo. But adding it to release pages and CDN is a great idea.

simplePCBuilding:
Okay. I think I cannot add anything to the releases, right? Also, what do you think about the latest changes?

simplePCBuilding:
Okay. I think I cannot add a anything to the releases, right? Also, what do you think about the latest changes?

simplePCBuilding:
LOL that was weird. I commented on my phone which at school has really bad connection, GitHub spat out a network error, but my comment got actually published.

GettingStarted.md
```diff
@@ -0,0 +1,162 @@
+# Introduction
+Welcome to impress.js! This presentation framework allows you to create stunning presentations with the power of CSS3 transformations. It is inspired by the idea behind [prezi.com](https://prezi.com). 
+**NOTE:** This Guide is not made for you, if you have never written HTML and/or CSS before. Knowing your way around in JavaScript certainly helps, but is not a necessity. You may still continue this tutorial and try to understand what we do as you go. 
+
+Also, if you are an experienced (or sorta experienced user, anyway), you might prefer the [DOCUMENTATION](DOCUMENTATION.md).
```
henrikingo, line 5: I'm sorry to be a language nitpick...This kind of casual language can be nice in a way, but also it makes the text longer without adding information value. I would prefer: 

    For more advanced and complete documentation, see [DOCUMENTATION](DOCUMENTATION.md)

GettingStarted.md
```diff
@@ -0,0 +1,162 @@
...
+# Getting started with impress.js
+## Installation / acquiring the framework
+You first need to decide if you want to use the bleeding edge (most up to date, generally very reliable as well, but may contain more problems) or the stable release. You can use one of the methods mentioned below or the direct download link to only the javascript file. The benefit is, you don't have to find the file in the repository and you can copy the link and directly load the file from the cdn. 
+
+### **Direct links to only impress.js file**
+- V2.0.0: https://cdn.jsdelivr.net/gh/impress/impress.js@2.0.0/js/impress.js
+- V1.1.0: https://cdn.jsdelivr.net/gh/impress/impress.js@1.1.0/js/impress.js
+- Source: https://cdn.jsdelivr.net/gh/impress/impress.js/js/impress.js
+
+For older versions, please just replace the version number behind the @!
```
henrikingo, line 16: I feel like this sentence can be removed. It's useful, but in reality very few people will ever need to use a specific older version. So it's not something that needs to be in a simple Getting Started guide.

GettingStarted.md
henrikingo: In this section, can you just use the code above. It just works, no need to explain or understand why.

GettingStarted.md
```diff
@@ -0,0 +1,162 @@
...
+
+### **Stable release:**
+Head to the releases tab and download the source code as zip or as a tarball. Go ahead and unzip / untar it. You may find the framework under ```/js/impress.js```. Do also include the ```/css/impress-common.css```, if you want to make your life a bit easier.
+
+### **Bleeding Edge:**
+In the github repository, click the "Code"-Button, then click "Download ZIP". Go ahead and unzip / untar it. You may find the framework under ```/js/impress.js```. Do also include the ```/css/impress-common.css```, if you want to make your life a bit easier.
+
+## Setting up the project
+Open up your favorite text-editor / IDE, for example Visual Studio Code, Atom, Notepad ++, ...
+Now, copy the *impress.js* file into the folder you are working in and create a new file called *index.html* and create the basic HTML layout:
```
henrikingo, line 40: While everything you say is correct technically, it's important to be consistent. In all demo presentations and other guidance, even within this file just a few lines above, the `impress.js` file is in a directory called `js/`. So you should keep that through all examples:

    Copy the folder `js/` into the folder you are working in. It contains the `impress.js` file.

GettingStarted.md
```diff
@@ -0,0 +1,162 @@
...
+-------------------------|-----------|------------
+data-transition-duration | 1000 (ms) | Speed of transition between steps.
+data-width               | 1920 (px) | Width of target screen size. When presentation is viewed on a larger or smaller screen, impress.js will scale all content automatically.
+data-height              | 1080 (px) | Height of target screen size.
+data-max-scale           | 3         | Maximum scale factor. (Note that the default 1 will not increase content size on larger screens!)
+data-min-scale           | 0         | Minimum scale factor.
+data-perspective         | 1000      | Perspective for 3D rendering. See https://developer.mozilla.org/en/CSS/perspective
+
+### **Renaming Steps**
+You can give each step an ID. The name of the ID will be displayed in the browsers navigation bar instead of the default *step-x* whereas x is replaced by the current step number. This can be especially helpful, when trying to jump between steps and go back to a previous one. If you want to know how to move to a specific slide, you should take a look at the [DOCUMENTATION](DOCUMENTATION.md). This functionality is part of the "Goto" plugin
```
henrikingo, line 132: This is not true that plugins would be documented in DCOUMENTATION.md

GettingStarted.md
```diff
@@ -0,0 +1,162 @@
...
+
+impress.js has accumulated a lot of very useful plugins. Read the [GettingStarted](/src/plugins/GettingStarted.md)-Guide in the plugins folder to learn everything in more depth.
+
+Each Plugin has a README.md file which you may read to get an idea on how to use them. Some of the plugins run unnoticed in the background, like the *resize* plugin, which automatically resizes the presentation whenever the browser window changed in size. Here, I will give you an overview of some of the plugins that impress.js includes by default. 
+
+**NOTE:** As previously mentioned, if you'd like to get more info about how it works, take a look at the [DOCUMENTATION](DOCUMENTATION.md) or the README.md files of the plugins.
+
+## [impressConsole](/src/plugins/impressConsole/README.md)
+This plugin opens up and additional browser tab which contains a speaker console. There you can see the current slide, the past slide and your notes. You add notes to your presentation by adding a *div* that belongs to the class "notes" to your *div* that belongs to the class "step". 
+
```
henrikingo, line 145: Please add example of using notes

README.md
```diff
@@ -13,6 +13,16 @@ impress.js may not help you if you have nothing interesting to say ;)
...
+### Getting Started Guide
+Check out our new [Getting Started](GettingStarted.md) guide if you want a quick introduction to the project!
+
+### Direct download link to only impress.js
+You can include this link directly inside of your HTML file in its header. If you want to learn how to do this, you can find a how-to in the [Getting Started](GettingStarted.md) guide.
+- V2.0.0: https://cdn.jsdelivr.net/gh/impress/impress.js@2.0.0/js/impress.js
+- V1.1.0: https://cdn.jsdelivr.net/gh/impress/impress.js@1.1.0/js/impress.js
+- Source: https://cdn.jsdelivr.net/gh/impress/impress.js/js/impress.js
+
+For older versions, please just replace the version number behind the @!
```
henrikingo, line 25: Again, can be removed.

GettingStarted.md
```diff
@@ -0,0 +1,162 @@
+# Introduction
+Welcome to impress.js! This presentation framework allows you to create stunning presentations with the power of CSS3 transformations. It is inspired by the idea behind [prezi.com](https://prezi.com).
```
henrikingo, line 2: I would leave out the prezi.com attribution. impress.js is so much more at this point.

GettingStarted.md
```diff
@@ -0,0 +1,162 @@
+# Introduction
+Welcome to impress.js! This presentation framework allows you to create stunning presentations with the power of CSS3 transformations. It is inspired by the idea behind [prezi.com](https://prezi.com). 
+**NOTE:** This Guide is not made for you, if you have never written HTML and/or CSS before. Knowing your way around in JavaScript certainly helps, but is not a necessity. You may still continue this tutorial and try to understand what we do as you go. 
+
+Also, if you are an experienced (or sorta experienced user, anyway), you might prefer the [DOCUMENTATION](DOCUMENTATION.md).
+
+# Getting started with impress.js
+## Installation / acquiring the framework
+You first need to decide if you want to use the bleeding edge (most up to date, generally very reliable as well, but may contain more problems) or the stable release. You can use one of the methods mentioned below or the direct download link to only the javascript file. The benefit is, you don't have to find the file in the repository and you can copy the link and directly load the file from the cdn. 
+
```
henrikingo, line 10: For a getting started guide, keep it simple and avoid asking the user to decide. YOU should decide. Just show one option. Which one would you prefer?

henrikingo (commented):
Some more comments on the text

henrikingo:
> Okay. I think I cannot add anything to the releases, right? 

Ah right. Can you file a new ticket and me or someone else could look into it in the future.

simplePCBuilding:
I have updated my repo to include the changes you mentioned

GettingStarted.md
```diff
@@ -129,7 +126,7 @@ data-min-scale           | 0         | Minimum scale factor.
 data-perspective         | 1000      | Perspective for 3D rendering. See https://developer.mozilla.org/en/CSS/perspective
 
 ### **Renaming Steps**
-You can give each step an ID. The name of the ID will be displayed in the browsers navigation bar instead of the default *step-x* whereas x is replaced by the current step number. This can be especially helpful, when trying to jump between steps and go back to a previous one. If you want to know how to move to a specific slide, you should take a look at the [DOCUMENTATION](DOCUMENTATION.md). This functionality is part of the "Goto" plugin
+You can give each step an ID. The name of the ID will be displayed in the browsers navigation bar instead of the default *step-x* whereas x is replaced by the current step number. This can be especially helpful, when trying to jump between steps and go back to a previous one. If you want to know how to move to a specific slide, you should take a look at the [DOCUMENTATION](./src/plugins/goto/README.md) of the "Goto" plugin.
```
henrikingo, line 129: Strictly speaking it's README, not DOCUMENTATION. Again, it's not a big deal semantically, but can be confusing to areader.

README.md
```diff
@@ -13,6 +13,16 @@ impress.js may not help you if you have nothing interesting to say ;)
 
 HOW TO USE IT
 ---------------
+### Getting Started Guide
+Check out our new [Getting Started](GettingStarted.md) guide if you want a quick introduction to the project!
```
henrikingo, line 17: But this is the GettingStarted guide? Why advertise yourself?

README.md
henrikingo: ...The last line.

henrikingo (commented):
Thanks. You missed a couple and I added 2 more comments.

GettingStarted.md
simplePCBuilding: Fair enough, will resolve

README.md
simplePCBuilding: Didn't remember I did put this into the README as well lol. Will fix

README.md
simplePCBuilding: This is in the README.md file

simplePCBuilding:
I saw it, thank you

README.md
henrikingo: Oops, my bad

simplePCBuilding:
I just checked again that I didn't miss any of the previously mentioned things, I didn't

GettingStarted.md
```diff
@@ -0,0 +1,178 @@
...
+
+        <div id="impress">
+            <div id="myFirstSlide" class="step">
+                <h1>My first Slide</h1>
+            </div>
+        </div>
+
+        <script src="https://cdn.jsdelivr.net/gh/impress/impress.js@2.0.0/js/impress.js"></script><!--Only use, if you have internet connection when giving the presentation!-->
+
+        <script type="text/javascript" src='./js/impress.js></script><!--Only use, if you are not using the cdn!-->
```
henrikingo, line 60: Okay, last question...

Have you tested what happens if someone copies this as is, and keeps both of the script tags above? (And both files exist,)

simplePCBuilding: I have not, but I will do. I am pretty sure I already did something like that by accident with I believe jQuery but not with impress. It still worked

simplePCBuilding: I've checked on my iPad and it is working. I will check with LibreWolf as well (which is just Firefox with more privacy)

simplePCBuilding: I have a suggestion to root out any possibility of something bad happening. I will just comment one of them out and tell the user to just uncomment it.

simplePCBuilding: Okay, it caused problems, it skips steps (so only every other step is being displayed)

henrikingo: Interesting. That's because it would load both of them, so 2 copies of impress both move forward 1 step, at the same moment.

Thanks for testing.

My suggestion would be to go back to this:

    <script> src="js/impress.js"></script>
    <script>window.impress || document.write('<script src="https://cdn.jsdelivr.net/gh/impress/impress.js@2.0.0/js/impress.js">\x3C/script>');</script>

Although it's a bit unreadable to a beginner, it has the benefit that it just works even if you don't understand it.

(I modified the above compared to previous discussion so that it first tries the local copy, and the remote is the fallback.)

simplePCBuilding: Yeah. It is. I assumed browser cache would get it, but this isn't the case. I will change it to what you suggested

simplePCBuilding: Ohh I just noticed a problem with the CDN... You need to have a server like apache, http-server (which is a node module), as network requests are not allowed in the file protocol afaik.

simplePCBuilding: oh nvm

simplePCBuilding:
I have also bumped the version number to 2.1.0 in the impress.js file. It was still on V1.1.0. I might be wrong with that version number, but it'd make sense, as 2.0.0 is already out.

simplePCBuilding:
Also, I just noticed, that the copyright info is really outdated in the README.md. Shall I update it, or will somebody else do it?

GettingStarted.md
```diff
@@ -0,0 +1,167 @@
...
+Loading the script from the cdn is quite straight forward. If you copy the below example code, you need to do nothing else, impress will be loaded automatically.
+
+**Direct links to different versions of the impress.js file**
+- V2.0.0: https://cdn.jsdelivr.net/gh/impress/impress.js@2.0.0/js/impress.js
+- V1.1.0: https://cdn.jsdelivr.net/gh/impress/impress.js@1.1.0/js/impress.js
+- Source: https://cdn.jsdelivr.net/gh/impress/impress.js/js/impress.js
+
+## Download the file to your PC
+Head to the releases tab and download the source code as zip or as a tarball. Go ahead and unzip / untar it. You need to copy the folder */js/* and the folder */css/* (if you want to make your life a bit easier) into the folder you are working in. 
+
```
henrikingo, line 21: I still think this is a bit fuzzy. It's not clear that you don't *need* to download /css/* rather that is **entirely** optional.

My suggestion would be to just mention `/js/*` in this section, and then mention `/css/*` only in the section further below that you start to speak about CSS.

GettingStarted.md
```diff
@@ -0,0 +1,167 @@
...
+data-rotate-y   | Rotate the element along the y-axis
+data-rotate-z   | Rotate the element along the z-axis
+data-scale      | Scale the element. 
+
+These are the basic positioning options in impress.js. All of the attributes take Strings as arguments, so be aware of the fact, that you need to put quotation marks around the numbers! The *data-rotate* attributes take an angle in form of a String as argument.
+
+Now, that you have created the slides, you might want to style them. This is where CSS comes into play. Add another file to your project called, e.g., ```style.css```. 
+
+**NOTE:** As a general rule, you should avoid modifying the CSS properties of the built-in class *step*, as this will modify ***ALL*** the steps! You may change the font of it, but I recommend to instead add another class to that element and style that class. Whatever you do, do not mess with positioning and rotation of the div that belongs to the class *step*, but add a div inside of it, if you really have to mess with these properties. See the example below. Always position *steps* with the *data-* attribute!
+
```
henrikingo, line 95: Sorry... I've thought about this some more, and I realize now this just isn't right.

We should think about this from the point of view that `impress-common.css` is *optional*. And if a user wouldn't use it, they rather MUST add CSS to the `.step` class. So the conclusion is that a user can and sometimes MUST add CSS to the `.step` class.

I would simply delete the first half of this section.But keep the latter half, from "Whatever you do..." to the end.

src/impress.js
```diff
@@ -6,13 +6,13 @@
  * in modern browsers and inspired by the idea behind prezi.com.
  *
  *
- * Copyright 2011-2012 Bartek Szopka (@bartaz), 2016-2020 Henrik Ingo (@henrikingo)
+ * Copyright 2011-2012 Bartek Szopka (@bartaz), 2016-2023 Henrik Ingo (@henrikingo)
```
henrikingo, line 9: Thanks!

I was thinking, let's change this to...

   * Copyright 2011-2012 Bartek Szopka (@bartaz), 2016-2023 Henrik Ingo (@henrikingo) and [70+  other contributors](https://github.com/impress/impress.js/graphs/contributors)

And yes, please update the README file(s) too

src/impress.js
```diff
@@ -6,13 +6,13 @@
...
  *
  * Released under the MIT License.
  *
  * ------------------------------------------------
  *  author:  Bartek Szopka, Henrik Ingo
- *  version: 1.1.0
+ *  version: 2.1.0
```
henrikingo, line 15: Ah yes... You're right that this is how most projects would do it. To be honest here I have kept the number that is the latest release, and then increase at release time, if I remember...

So with that logic this should be 

    2.0.0

I don't feel strongly about this and if anything I'd want this to be scripted instead of manually updated.

henrikingo (changes requested):
Thanks. This is really good now.

Sorry, I'm going back and forth I guess, but I have to ask you to change the part about CSS some more. Other than that I'm quite satisfied.

GettingStarted.md
simplePCBuilding: Was also missing another # there, as this was supposed to be h3 class

simplePCBuilding:
fixed everything you mentioned. Also I think it is important that this has a high standard

henrikingo:
> fixed everything you mentioned. Also I think it is important that this has a high standard

Yes, I appreciate your dedication and patience.

It's not so much that I hesitate to strive for perfection... I just feel bad if you've waited for review 4-6 weeks and then I ask to change one more thing.

But now this is done! Thanks so much for doing this!

PR #834, opened 2022-12-12
Bugfix for notes not being hidden reliably

simplePCBuilding:
I noticed that notes are not hidden properly all the time. I have read through the source code and have actually not found any code that would hide the notes.


simplePCBuilding:
I have now also removed the getting started guide from this PR

README.md
```diff
@@ -13,6 +13,8 @@ impress.js may not help you if you have nothing interesting to say ;)
 
 HOW TO USE IT
 ---------------
+### Getting Started Guide
```
fnogatz, line 16: This probably shouldn't be checked into this PR.

fnogatz (changes requested):

README.md
simplePCBuilding: Yeah right. Will remove this

simplePCBuilding:
Should be fixed now

simplePCBuilding:
Let's ignore that I just removed one too many lines

henrikingo:
Okay so there are notes in this presentation, and they seem to work. Can you confirm?

https://impress.js.org/examples/classic-slides/

simplePCBuilding:
Yeah it is hidden from the CSS file. I felt like it should be hidden from JS.

henrikingo:
This is another thing that should be in impress-common.css. But definitely CSS, not JS.

fnogatz:
I agree with @henrikingo here. In particular, adding `display: none` directly to the element's style makes it unnecessarily hard in case you *actually* want to show notes. For instance, what I really like about Google's *io-2012-slides* is that they allow you not only to see the presenter notes in the presenter mode, but also by pressing the hotkey 'P' anywhere in the presentation. See for instance [this example](https://www.dfki.uni-kl.de/~nogatz/DFKI-HTML-Slides/index.html#10) and press 'P'. However, this would require a `display: block !important` when hiding is done by JavaScript instead of just using CSS.

simplePCBuilding:
I see that, I'll close the PR

PR #841, opened 2022-12-26
Highlight future substeps in slideView

fnogatz:
In the current version of impress.js, the *slideView* in the impress-console is an exact copy of the slide that is currently presented. In my opinion, this ignores one essential benefit of this view, namely to additionally give the presenter hints on what is coming next.

This PR adds a CSS snippet that indicates the next *substeps* by a red border:

Screenshot from 2022-12-26 14-45-43

In addition, if at least one substep has the `data-substep-order` attribute set, the corresponding step number is displayed as well:

Screenshot from 2022-12-26 14-46-53

The highlighting is removed once the substep is displayed:

Screenshot from 2022-12-26 14-48-15

This PR is a follow-up on #825 and #827.

### DIFF
--- a/css/impress-common.css
+++ b/css/impress-common.css
@@ -111,6 +111,42 @@ It is focused on plugin functionality, not the visual style of your presentation
 }
 
 /*
+  Highlight substeps that are not yet visible in the sliveView window of the
+  impressConsole.
+*/
+.impress-console.slideView #impress .step:not(.future) .substep:not(.substep-visible) {
+  position: relative;
+  opacity: 1.0;
+}
+.impress-console.slideView #impress .step:not(.future) .substep:not(.substep-visible):before {
+  position: absolute;
+  top: 0;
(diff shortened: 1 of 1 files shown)


henrikingo:
This is a good idea, but I suggest a different implementation would be more elegant:

Change the logic of the preView window so that if there are substeps, it will show the next substep, not next slide.

fnogatz:
Hm, but this way there's still no view that shows the *final* state of this slide. The highlighting not only serves the purpose of showing what will appear in the very next substep. Personally, I often need a reminder like "Don't talk about this point now, it will come later" :sweat_smile:

henrikingo:
Ok fair point. So I would still argue that what you're proposing should be the preView window, and the slideView should be as similar to the real/main window as possible.

Once all the substeps are visible, then the slideView window should contain the current step with all substeps visible, and the preView window should show the next step.

PR #851, opened 2023-03-05
replace event.keyCode with event.code, tweak documentation

### LINKED ISSUE #849: event.keyCode is deprecated, should upgrade to event.code
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode states
> You should avoid using this if possible; it's been deprecated for some time. Instead, you should use [KeyboardEvent.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code), if it's implemented. Unfortunately, some browsers still don't have it, so you'll have to be careful to make sure you use one which is supported on all target browsers.

`event.keyCode` shows up in places like https://github.com/impress/impress.js/blob/9475720f42f8b8ffcf6efafe2f5bf80a9e59fcbd/src/plugins/navigation/navigation.js#L95-L107

mengwong:
resolves https://github.com/impress/impress.js/issues/849

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,6 @@ You can include this link directly inside of your HTML file in its header. If yo
 - V1.1.0: https://cdn.jsdelivr.net/gh/impress/impress.js@1.1.0/js/impress.js
 - Source: https://cdn.jsdelivr.net/gh/impress/impress.js/js/impress.js
 
-### Getting Started Guide
-Check out our new [Getting Started](GettingStarted.md) guide if you want a quick introduction to the project!
-
 ### Checking out and initializing the git repository
 
     git clone --recursive https://github.com/impress/impress.js.git
@@ -82,7 +79,7 @@ REPOSITORY STRUCTURE
 * [test/](test/): Contains QUnit and Syn libraries that we use for writing tests, as well as some test coverage for core functionality. (Yes, more tests are much welcome.) Tests for plugins are in the directory of each plugin.
 * [js/](js/): Contains [js/impress.js](js/impress.js), which contains a concatenation of the core `src/impress.js` and all the plugins. Traditionally this is the file that you'll link to in a browser. In fact both the demo and test files do exactly that.
 * [css/](css/): Contains a CSS file used by the demo. This file is **not required for using impress.js** in your own presentations. Impress.js creates the CSS it needs dynamically.
-* [extras/](extras/) contains plugins that for various reasons aren't enabled by default. You have to explicitly add them with their own `script` element to use them.
+* [extras/](https://github.com/impress/impress-extras/) contains plugins that for various reasons aren't enabled by default. You have to explicitly add them with their own `script` element to use them. (This is provided as a Git submodule so if this directory is empty, you probably didn't clone with `--recursive`. Try running `git submodule update --init --recursive --remote` to get it.)
 * [build.js](build.js): Simple build file that creates `js/impress.js`. It also creates a minified version `impress.min.js`, but that one is not included in the github repository.
 * [package.json](build.js): An NPM package specification. This was mainly added so you can easily install [buildify](https://www.npmjs.com/package/buildify) and run `node build.js`. Other than the build process, which is really just doing roughly `cat src/impress.js src/plugins/*/*.js > js/impress.js`, and testing, `impress.js` itself doesn't depend on Node or any NPM modules.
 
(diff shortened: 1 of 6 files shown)


mengwong:
sigh, fighting with linter, will push again in a moment.

README.md
```diff
@@ -22,9 +22,6 @@ You can include this link directly inside of your HTML file in its header. If yo
 - V1.1.0: https://cdn.jsdelivr.net/gh/impress/impress.js@1.1.0/js/impress.js
 - Source: https://cdn.jsdelivr.net/gh/impress/impress.js/js/impress.js
 
-### Getting Started Guide
```
simplePCBuilding, line 25: Thank you, did not notice that whilst merging the second one ended up in there as well

README.md
```diff
@@ -82,7 +79,7 @@ REPOSITORY STRUCTURE
...
 * [js/](js/): Contains [js/impress.js](js/impress.js), which contains a concatenation of the core `src/impress.js` and all the plugins. Traditionally this is the file that you'll link to in a browser. In fact both the demo and test files do exactly that.
 * [css/](css/): Contains a CSS file used by the demo. This file is **not required for using impress.js** in your own presentations. Impress.js creates the CSS it needs dynamically.
-* [extras/](extras/) contains plugins that for various reasons aren't enabled by default. You have to explicitly add them with their own `script` element to use them.
+* [extras/](https://github.com/impress/impress-extras/) contains plugins that for various reasons aren't enabled by default. You have to explicitly add them with their own `script` element to use them. (This is provided as a Git submodule so if this directory is empty, you probably didn't clone with `--recursive`. Try running `git submodule update --init --recursive --remote` to get it.)
```
simplePCBuilding, line 82: Good one. Some people might be confused by that if they've not worked with git that much before

simplePCBuilding:
Thank you for doing this. It's important to keep things up to date with the newest spec of ES6

mengwong:
> Thank you for doing this. It's important to keep things up to date with the newest spec of ES6. When was this feature introduced? I'm tho kinda worried about impress stopping to work in older browsers. If this is a breaking change, please add a note to it in the README.md, the GettingStarted.md guide and the DOCUMENTATION.md files so people are informed about it.

It looks like all browsers added support in the 2015–2017 era, the last holdout was Edge in 2020. Would you consider it a breaking change?

janishutz (changes requested):
There are errors in the test I don't quite understand, please look into it. I assume you might have missed a key-event and that's why it does not complete the tests, but I am not sure. https://app.circleci.com/pipelines/github/impress/impress.js/210/workflows/011fdf79-48e8-4a0f-a2ba-913b55905d9a/jobs/441

mengwong:
I am trying really hard to reproduce this, I don't understand it either. I am sure it will be facepalmingly obvious once it is solved.

simplePCBuilding:
What I expected to be a problem with this solution: https://caniuse.com/?search=keyboardevent.code

This will break impress.js in older browsers like IE and mobile browsers (aka. Safari for iOS, Firefox for Android, etc.)

Compare it to: https://caniuse.com/?search=keyboardevent.keyCode, which is supported by almost every major browser out there. 

We should consider implementing a way to notify the user about incompatible plugins & automatically disable the plugins relying on these features, if the browser does not support it to retain support for old browsers with the most recent versions of impress.js.

Although this is great to offer, I personally am in favour of dropping support for the old browsers barely anyone uses any more nowadays. People that need to use those browsers should use a different CDN-Link or V2.0.0 instead of a new (future) release or upstream.

simplePCBuilding:
> I am trying really hard to reproduce this, I don't understand it either. I am sure it will be facepalmingly obvious once it is solved.

That might happen. That's just life when developing stuff lol. Randomly troubleshooting problems that do not exist is still worse though imo

mengwong:
Re support for old browsers I feel like this project in particular should be able to afford upgrades since in most cases presentations tend to be ephemeral and are delivered by authors who are in control of both the browser and the npm deps. Presentations that need to be archival-quality could pin to a release version but the risk of bitrot tends to run in the opposite direction, so staying current is better.

simplePCBuilding:
Exactly

simplePCBuilding:
@henrikingo ready to merge, should work now

henrikingo:
Oh, apologies, I've been ignoring this PR as the checks are still shown as failed. If tests really are passing for you, then could you please trigger a new CircleCI build so that it is shown on this PR as well? You can do that easily by pushing a new commit with just a whitespace change somewhere.

henrikingo:
Note to self, CI moved from CircleCI to Github Actions. Need to rebase this to confirm it still passes tests. Then ready to merge.

Issue #850, opened 2023-03-04
Should the goto plugin support non-navigation input keys?


mengwong:
Currently, the goto plugin recognizes keys like `ArrowUp`, `ArrowDown` in `data-goto-key-list`.

Wouldn't it be nice if the goto plugin could recognize keys like `KeyA`, `KeyB`, `Digit1`, `Digit2` as well?

Unless there is a good reason not to do this, I will fork and work toward submitting a PR to support this feature.

simplePCBuilding:
Hhm. What would be the reason for it? I am pretty sure space bar is already in the list. Would you then in that case map the keystroke to a specific step with something like  
```
data-goto-step-key="A"
```
?

mengwong:
yes, I will PR a working demo. I thought about it a bit more and realized I was implementing bookmark hotkeys, so I have done it as a `bookmark` plugin, extending the existing 2d navigation.

simplePCBuilding:
I think you can close this issue now, as you have opened the PR, same with the other one

PR #852, opened 2023-03-05
New "bookmark" plugin allows hotkey fast-travel to specific steps

mengwong:
resolves https://github.com/impress/impress.js/issues/850


mengwong:
i will make the linter happy

simplePCBuilding:
You still have double quotes in the code that's why it complains. Please run 
```
npm run all
```

before pushing

src/plugins/bookmark/bookmark.js
```diff
@@ -18,48 +22,51 @@
...
+	// window.console.log(event)
         if ( ( !event || !event.origEvent ) ) { return; }
-	var dest = hotkeyDest(event.origEvent)
-	// window.console.log(`bookmark: recognizing hotkey ${event.origEvent.code} goes to ${dest}`)
+	var dest = hotkeyDest( event.origEvent );
         if ( dest ) {
-	    // window.console.log(`bookmark: recognizing hotkey ${event.code} goes to ${dest}`)
-            var newTarget = document.getElementById( dest[0] ); // jshint ignore:line
+
+	    // Window.console.log(`bookmark: recognizing hotkey ${event.code} goes to ${dest}`)
```
simplePCBuilding, line 62: Do you really want to leave the debug code commented out in here?

src/plugins/bookmark/bookmark.js
```diff
@@ -18,48 +22,51 @@
...
-            if ( div.dataset.bookmarkKeyList !== undefined && div.id != undefined) {
-		div.dataset.bookmarkKeyList.split( " " ).forEach( (k) => {
-		    if (hotkeys.hasOwnProperty(k)) hotkeys[k].push( div.id ); // if hotkeys collide, we'll cycle through the targets
-		    else hotkeys[k] = [div.id]; } ); } });
+	root.querySelectorAll( ".step" ).forEach( function( div ) {
+            if ( div.dataset.bookmarkKeyList !== undefined && div.id !== undefined ) {
+		div.dataset.bookmarkKeyList.split( " " ).forEach( ( k ) => {
+		    if ( hotkeys.hasOwnProperty( k ) ) {
+			hotkeys[ k ].push( div.id );
+		    } else { hotkeys[ k ] = [ div.id ]; } } ); } } );
```
simplePCBuilding, line 41: Can you move those closing brackets to separate lines? This makes the code more legible for beginners, imo. Also some explanation might be needed for the code here, if someone with less coding (well js, anyway) knowledge than you have reads this code.

src/plugins/bookmark/bookmark.js
```diff
@@ -18,48 +22,51 @@
  * Copyright 2016-2017 Henrik Ingo (@henrikingo)
```
simplePCBuilding, line 22: You can credit yourself here as well!

src/plugins/bookmark/README.md
```diff
@@ -0,0 +1,27 @@
...
+```html
+<!-- data-bookmark-key-list allows an "inbound"-oriented style of non-linear navigation. -->
+<div id="..." class="step" data-bookmark-key-list="Digit1 KeyA 1 2 3 a b c">
+```
+
+An `id` is required on the `div`.
+
+If you assign the same key to multiple steps, that hotkey will cycle among them.
+
+WARNING: It's up to you to avoid reserved hotkeys H, B, P, ?, etc.
```
simplePCBuilding, line 20: Couldn't you just disallow the usage of those keys?

simplePCBuilding (changes requested):
Thank you for creating this plugin. My suggestion is to add some more in-code documentation, so it is easier for a new user to read the code and understand what happens

src/plugins/bookmark/README.md
mengwong: Don't think so, it would clobber the other plugins … I'll think about how to resolve the interaction effects.

mengwong:
You know, the weird thing that's going on, is that both my PRs, which are generally unrelated, both are failing tests in similar ways … maybe a curse is following me around

src/plugins/bookmark/bookmark.js
mengwong: lol i'm surprised the linter even let me do that … haha

I don't really think there's anything to be gained from adding a bunch of punctuation over multiple lines…  my preference is to keep things as short as possible so you can see everything on one screen. They used to say "you can write Fortran in any language", now I suppose they should say "you can write Python in any language" 😺

mengwong:
if I could get this to run after init, the `step-N` ids would be allocated, and I wouldn't have to worry about divs not having ids.

src/plugins/bookmark/bookmark.js
mengwong: i have taken it out, will push a new rev. Still working on the test failure.

mengwong:
so, I did run `npm run all` before submitting the PR, and it turns out that fixing the linter errors was what led to the test failures. Everything is fine at 2c43ff3 😄

mengwong:
ah, null !=== undefined

https://www.quora.com/Why-did-Tony-Hoare-the-developer-of-Android-Studio-says-The-Null-References-is-my-billion-dollar-mistake

simplePCBuilding:
Fun fact, this is actually one of my most favourite (and at the same time most hated) features of js

mengwong:
Thanks for helping me get this PR shipshape, I appreciate it. I am building an app involving this codebase and will keep an eye out for problems. If there are no problems I will follow up in a while and maybe we can have another think about integrating this into master.

simplePCBuilding:
Okay, great! Can you post a link to your project here, if it's open source?

mengwong:
> Okay, great! Can you post a link to your project here, if it's open source?

I will … once it is more presentable … still trying to figure out a bunch of things.

src/plugins/bookmark/bookmark.js
simplePCBuilding: The linter is not properly configured for that. I have an issue open to talk about moving to eslint and how to configure it

src/plugins/bookmark/README.md
henrikingo: I think this is ok. If you want to override a reserved key, just do it.

Issue #855, opened 2023-03-18
Default substep order makes all substeps appear at once


treitmayr:
The [Substep Plugin documentation](https://github.com/impress/impress.js/blob/4c9d2e38aaf1ea038118d4ff6549609df6f1fab3/src/plugins/substep/README.md) states for the default order of substeps:

> By default, this plugin reveals substeps in the order in which they appear in the HTML.

However this does not seem to work correctly - when no order is specified, then all substeps appear at once when only the first substep should be visible. This behavior was probably introduced with commit 4c9d2e38aaf1ea038118d4ff6549609df6f1fab3, but I did not test it before that, to be honest.
Both going to the next and to the previous substep are affected.

There seems to be a simple fix, which I will provide as a PR.

treitmayr:
Closing issue because PR #856 is going to be merged.

PR #856, opened 2023-03-18
Fix default substep order when no explicit order given (#855)

treitmayr:
This PR fixes the default order of substeps, for which no explicit order was given (see issue #855). It ensures that the loop for marking substeps "active" + "visible" is left when at least one substep (with explicit or default order) was marked and this substep uses default order.

Note: The removed space in the header of js/impress.js does not relate to the current issue, but comes from regenerating this file. Apparently the file has not been regenerated when its source file was modified in some previous commit.

### DIFF
--- a/js/impress.js
+++ b/js/impress.js
@@ -8,7 +8,7 @@
  * in modern browsers and inspired by the idea behind prezi.com.
  *
  *
- * Copyright 2011-2012 Bartek Szopka (@bartaz), 2016-2023 Henrik Ingo (@henrikingo) 
+ * Copyright 2011-2012 Bartek Szopka (@bartaz), 2016-2023 Henrik Ingo (@henrikingo)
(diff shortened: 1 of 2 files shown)


treitmayr:
I had to update the PR because I realized that not only going to the next substep was broken, but also going to the previous one. Should work now.

fnogatz:
Thank you, @treitmayr, for your contribution! Indeed, this was a bug in the substep plugin. I have not looked in detail into its history, but maybe it was introduced by my modifications in https://github.com/impress/impress.js/pull/825 :see_no_evil: 

I tested your modifications and they work as expected, thank you! @henrikingo, this is good to merge :)

simplePCBuilding (approved):
Works as it should, ready to merge

sivakumar2805 (approved):

Issue #860, opened 2023-06-16
Bug in GettingStarted.


MagnusMG:
I tried the Getting Started demo, and the code for index.html left me with a page that always showed the fallback message. After some head-scratching and comparison with the demo file I discovered that the `impress-demo.css` also has to be included:
````
<link href="css/impress-demo.css" rel="stylesheet" />		
`````

simplePCBuilding:
I see what you mean. In the getting started guide I assumed that you also have the impress-common.css file. I‘ll come up with a fix and will open a PR possibly tomorrow with a fix for it

simplePCBuilding:
@MagnusMG Sorry for the delay, PR #861 is open now with the changes. I'd be happy if you could review the changes and give suggestions if needed

simplePCBuilding:
I am still awaiting this PR to be merged, as well as another massive one.

henrikingo:
This seems to now be mentioned in GettingStarted.md

GraffL:
It seems to me like this issue has not really been resolved yet (at least I just stumbled over the exact same problem and it took me quite some time to realize what was going on): 
I followed the Getting Strated guide and ended up with a presentation that always shows the "browser not supported"-message.

I think there was some misunderstanding in the above discussion as @MagnusMG initially (correctly) reported the problem as "The example in Getting Started fails to include impress-**demo**.css", but the following discussion (and subsequent pull request) was then about impress-**common**.css (which is already included in the example).

So, in my opinion the example code in the Getting Started Guide should just also include
`<link rel="stylesheet" href="./css/impress-demo.css">`

Alternativly, one could move 
`.impress-supported .fallback-message {
    display: none;
}`
from impress-demo.css to impress-common.css (as well as maybe some additional basic styling).

------

PS: As a side note: I find the text on the option without having to download anything a bit misleading as, of course, this only applies to the impress.js-file. One still has to download the css-files (or write them oneself from scratch).

simplePCBuilding:
Thank you for pointing this out. This is on my list for the big update that is coming probably towards the end of the year. I had already opened a PR that was declined.

PR #861, opened 2023-07-06
Add introduction to usage of CSS in GettingStarted.md

simplePCBuilding:
As discussed in #860, here are some changes to make the guide more complete, with also including some notes on CSS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,6 @@ You can include this link directly inside of your HTML file in its header. If yo
 - V1.1.0: https://cdn.jsdelivr.net/gh/impress/impress.js@1.1.0/js/impress.js
 - Source: https://cdn.jsdelivr.net/gh/impress/impress.js/js/impress.js
 
-### Getting Started Guide
-Check out our new [Getting Started](GettingStarted.md) guide if you want a quick introduction to the project!
-
 ### Checking out and initializing the git repository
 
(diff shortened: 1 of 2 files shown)


simplePCBuilding:
I will be working on some more on impress.js starting in November / December this year as I will again have time to work on other projects. Expect some cool PRs (I've got a few ideas)

README.md
```diff
@@ -22,9 +22,6 @@ You can include this link directly inside of your HTML file in its header. If yo
 - V1.1.0: https://cdn.jsdelivr.net/gh/impress/impress.js@1.1.0/js/impress.js
 - Source: https://cdn.jsdelivr.net/gh/impress/impress.js/js/impress.js
 
-### Getting Started Guide
-Check out our new [Getting Started](GettingStarted.md) guide if you want a quick introduction to the project!
-
```
henrikingo, line 27: Why is this removed??

henrikingo (commented):
Sorry but I don't get this PR? Why add more text for such a specific case? 

The guide instructs to use impress-common.css. If yolu leave it out, of course there will be consequences.

Similarly for the fix, why is the fix to add back CSS that you already deleted? Why isn't the fix to remove the fallback message entirely? (For example.)

I think I will close this. Sorry. But I don't see that this is needed.

PR #871, opened 2024-02-22
Fix HTML syntax in classic slides example

fnogatz:
The current HTML in `examples/classic-slides/index.html` is invalid syntax. This PR fixes the two problems.

In 81b49d0ec250e4e4b7b49b381b26cdcf066447e2, I removed the link to the tutorial at all, since inline links are disabled by default since Mermaid.js v8.2.0: see 'click functionality is disabled' in their [release notes](https://github.com/mermaid-js/mermaid/releases/tag/8.2.0).

### DIFF
--- a/examples/classic-slides/index.html
+++ b/examples/classic-slides/index.html
@@ -325,7 +325,7 @@
           A(Support for<br />diagrams)
           B[Provided by<br />mermaid.js]
           C{Already<br />know<br />mermaid?}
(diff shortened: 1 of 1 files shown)


simplePCBuilding (approved):
Good catch, I am currently working on bringing a massive update to impress, but that'll still take a couple of months. Didn't catch that so thanks

henrikingo:
Thanks

PR #874, opened 2024-09-08
[WIP] Add three plugin

kdxcxs:
## Introduction

It's a plugin makes three.js works with impress.js, which allows like 3d models to be added in impress.js canvas with all other feat from three.js.

**This plugin is still in a very early stage, please do not merge.** It's a cool feat for me, but I'm not sure if it's suitable for impess.js, so I opened this pr for further discution. The features are not fully implemented yet, but I've wrote a simple example in `examples/three/index.html`.

## TODO

- [ ] Support to add 3d models by defining data sets.
- [ ] Add util functions to allow users to control the 3d models programmatically.

## Issues

- As it's depend on three.js and tween.js, I'm not sure if it's ok for impress.js.
- How should plugin add the three.js canvas overlay into the dom?
  - It's currently implemented by adding a div#impress-three-container manually, maybe better to insert the canvas in plugin initialization with the root dom?
- As it's depend on three.js and tween.js, I'm not sure if it's ok for impress.js.
- The three.js overlay is always on the top of impress canvas, this could be wrong in position.
  - We could try creating a separate canvas for each 3d models and move with dom.
- The animation is not fully synced with css animation.

## Ending

Here's still a lot not mentioned above, and I think here could be a lot changes to be made in impress.js, so I choose to contact you guys as early as possible to make it right. Or if this is plugin not what impress.js wanted, it's ok to reject and close this pr.

### DIFF
--- a/examples/three/index.html
+++ b/examples/three/index.html
@@ -1,0 +1,114 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>impress.js with Three.js Plugin</title>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/three.js/r134/three.min.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/tween.js/23.1.3/tween.umd.js"></script>
+    <style>
+        body, html {
+            margin: 0;
+            padding: 0;
+            height: 100%;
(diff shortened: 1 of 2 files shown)


janishutz:
Can you please mark this as a draft?

janishutz:
@kdxcxs I have just had a bit of a look at three.js. As far as I can tell, it uses the HTML Canvas element, if I saw that correctly, am I right? If so, using it with impress should be fairly easy. That wouldn't even really require a plugin, right?

PR #895, opened 2026-07-23
docs(toolbar): document element param instead of html

### LINKED ISSUE #771: Doc of toolbar needs update
Currently, the doc in https://github.com/impress/impress.js/blob/master/src/plugins/toolbar/toolbar.js says:

```
html : "<button>Click</button>", // The html to add.
```

But the code uses `element`, not `html`:

```
group.appendChild( e.detail.element );
```

Hashim1999164:
## Summary
- Update the toolbar plugin header docs to use `detail.element` (a DOM node), matching the implementation and `navigation-ui`.
- Remove the outdated `html` / `callback` / `impress:toolbar:added` guidance, since that event is never fired.

Fixes #771.

## Test plan
- [ ] Read `src/plugins/toolbar/toolbar.js` header and confirm it matches `appendChild` / `insertBefore` listeners
- [ ] Cross-check against `navigation-ui.js` toolbar usage

### DIFF
--- a/src/plugins/toolbar/toolbar.js
+++ b/src/plugins/toolbar/toolbar.js
@@ -43,14 +43,11 @@
  *
  *    { group : 1,                       // integer. Widgets with the same group are grouped inside
  *                                       // the same <span> element.
- *      html : "<button>Click</button>", // The html to add.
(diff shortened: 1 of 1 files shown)


henrikingo:
Thanks!

